PR #5, opened 2016-09-28
Implement a Firehose -> Stackdriver nozzle

cholick:
This MVP connects to the firehose and sends HTTPStartStop events to Stackdriver.

Non-goals to be implemented (or not) later:
- Error handling
- Filtering of event types
- More configuration
- Mapping from events to labels?
- Alternate GCP authentication


src/stackdriver-nozzle/firehose/client.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (fc *client) StartListening(nozzle logging.Logging) error {
+	cfClient := cfclient.NewClient(fc.cfConfig)
+
+	cachingClient := caching.NewCachingEmpty()
+
+	eventRouter := eventRouting.NewEventRouting(cachingClient, nozzle)
+
+	//err := eventRouter.SetupEventRouting("LogMessage,ValueMetric,HttpStartStop,CounterEvent,Error,ContainerMetric")
```
johnsonj: is there anyway to specify * here? I'm assuming this commented line is a helper for later when you add all event types

src/stackdriver-nozzle/main.go
```diff
@@ -0,0 +1,59 @@
...
+	if *debug {
+		println("Sending firehose to standard out")
+		err := client.StartListening(&dev.StdOut{})
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		println("Sending firehose to Stackdriver")
+		sdClient := stackdriver.NewClient(*projectID)
+		//sdClient.Post("hello world 5")
```
johnsonj: remove

src/stackdriver-nozzle/main.go
```diff
@@ -0,0 +1,59 @@
...
+				Default("false").
+				OverrideDefaultFromEnvar("SKIP_SSL_VALIDATION").
+				Bool()
+	projectID = kingpin.Flag("project-id", "gcp project id").
+			OverrideDefaultFromEnvar("PROJECT_ID").
+			String() //maybe we can get this from gcp env...? research
+)
+
+func main() {
+	//todo: pull in logging library...
```
johnsonj: is this for the logging of the nozzle?

src/stackdriver-nozzle/nozzle/nozzle.go
```diff
@@ -0,0 +1,21 @@
...
+
+type Nozzle struct {
+	StackdriverClient stackdriver.Client
+}
+
+func (n *Nozzle) Connect() bool {
+	return true
+}
+
+func (n *Nozzle) ShipEvents(event map[string]interface{}, _ string /* TODO research second string */) {
```
johnsonj: curious what you have planned for the second string? maybe remove it for now

src/stackdriver-nozzle/main.go
nrxus: We have straight up print statements in here right now, but we really should use a real logging library so we can filter them by debug/info/etc...

src/stackdriver-nozzle/nozzle/nozzle.go
nrxus: The interface specifies the second string - it turns out to be a log message, but we haven't implemented any events that actually use it yet.

src/stackdriver-nozzle/firehose/client.go
nrxus: It doesn't look like it from the source, but we can create the full list ourselves from here:

https://github.com/cloudfoundry/sonde-go/blob/master/events/envelope.pb.go#L61-L76

When we implement the config for filtering, it should become moot, I think?

PR #7, opened 2016-09-29
Send Firehose ValueMetric events to Stackdriver monitoring

nrxus:
MVP for monitoring. We're leaning on [custom metric auto-creation](https://cloud.google.com/monitoring/custom-metrics/creating-metrics#auto-creation) to create our custom metric descriptors based on the `name` in the `ValueMetric`.

TODO:
- Add more metric events
- Don't throw away the rest of the event data (we currently only use the `name` and `value`)
- Figure out units - if we want to use units from the event, we can't use auto-creation (and we would need to do some additional work around only creating metric descriptors as needed)

The interface that we're provided from `firehose-to-syslog` passes a `map[string]interface{}`, so we lose quite a bit of type safety here. It didn't matter for `firehose-to-syslog` since it's just being outputted as text, but we may want to use the underlying library instead if this becomes an issue.


src/stackdriver-nozzle/nozzle/nozzle_test.go
```diff
@@ -16,68 +16,102 @@ var _ = Describe("Nozzle", func() {
...
 })
 
 type MockStackdriverClient struct {
-	PostFn func(interface{}, map[string]string)
+	PostLogFn    func(interface{}, map[string]string)
+	PostMetricFn func(string, float64) error
 }
 
 func (m *MockStackdriverClient) Post(payload interface{}, labels map[string]string) {
-	if m.PostFn != nil {
```
johnsonj: should this function be PostLog?

src/stackdriver-nozzle/stackdriver/client.go
```diff
@@ -25,25 +36,75 @@ const (
...
+		Name: projectName,
+		TimeSeries: []*monitoringpb.TimeSeries{
+			{
+				Metric: &google_api.Metric{
+					Type:   metricType,
+					Labels: map[string]string{},
+				},
+				Points: []*monitoringpb.Point{
+					{
+						Interval: &monitoringpb.TimeInterval{
```
johnsonj: is this a required param? this isn't much of an interval :)

src/stackdriver-nozzle/stackdriver/client.go
```diff
@@ -25,25 +36,75 @@ const (
...
+							},
+						},
+						Value: &monitoringpb.TypedValue{
+							Value: &monitoringpb.TypedValue_DoubleValue{
+								DoubleValue: value,
+							},
+						},
+					},
+				},
+				//MetricKind: google_api.MetricDescriptor_CUMULATIVE,
```
johnsonj: remove

johnsonj:
Ah tricky! Is it useful to store the time call and use the value twice so
there's no chance of two different values?

New to Go so not sure what behavior it has here.

kejadlen:
Good point, we should've done that. We're about to use the CF envelope timestamp for this though, so we'll do the right thing there.

PR #10, opened 2016-09-30
Pr noaa sonde

nrxus:
Removing the firehose to syslog dependency. We will instead now use NOAA and SONDE directly to get type safety. 

This is the MVP and currently only handles value metric event. Every other type of event is sent as a log instead of checking if it should be a metric instead. 

We currently auto create the metric descriptors, we probably want to move away from this since we cannot set units, which is information that we get from firehose. 

The only label we save right now is event type, but we can discuss what else makes sense to save as a label.


src/stackdriver-nozzle/firehose/client.go
```diff
@@ -25,30 +30,38 @@ func NewClient(apiAddress, username, password string, skipSSLValidation bool) Cl
...
-
-	//err := eventRouter.SetupEventRouting("LogMessage,ValueMetric,HttpStartStop,CounterEvent,Error,ContainerMetric")
-	err := eventRouter.SetupEventRouting("HttpStartStop,ValueMetric")
-	if err != nil {
-		return err
-	}
+	refresher := CfClientTokenRefresh{cfClient: cfClient}
+	cfConsumer.SetIdleTimeout(time.Duration(30) * time.Second)
+	cfConsumer.RefreshTokenFrom(&refresher)
+	messages, errs := cfConsumer.Firehose("test", "")
```
johnsonj: are these needed? looks like they're the subscription id, auth token. Should they be taken as config options?

src/stackdriver-nozzle/firehose/client.go
```diff
@@ -25,30 +30,41 @@ func NewClient(apiAddress, username, password string, skipSSLValidation bool) Cl
...
-			IdleTimeoutSeconds:     30,
-			FirehoseSubscriptionID: "stackdriver-nozzle",
+	for {
+		select {
+		case envelope := <-messages:
+			err := fh.HandleEvent(envelope)
+			if err != nil {
+				return err
+			}
+		case err := <-errs:
```
johnsonj: does Firehose guarantee it puts an error message in this channel at some point? Eg on a normal exit?

src/stackdriver-nozzle/nozzle/nozzle.go
```diff
@@ -1,31 +1,28 @@
...
-		name := event["name"]
-		count := event["value"]
-		n.StackdriverClient.PostMetric(name.(string), count.(float64))
-
+func (n *Nozzle) HandleEvent(envelope *events.Envelope) error {
+	labels := map[string]string{
+		"event_type": envelope.GetEventType().String(),
+	}
+	switch envelope.GetEventType() {
+	case events.Envelope_ValueMetric:
```
johnsonj: great types here :) Go NOAA

src/stackdriver-nozzle/firehose/client.go
evandbrown: Looks like the Firehose retries five times:

> Whenever an error is encountered, the error will be sent down the error
channel and Firehose will attempt to reconnect up to 5 times.  After five
failed reconnection attempts, Firehose will give up and close the error and
Envelope channels.

What about something like this?

```
for envelope := range messages {
	err := fh.HandleEvent(envelope)
	if err != nil {
		return err
	}
}

return <-errs
```

evandbrown: The error channel has a buffer size of 1, so the above won't work. This should do it:

```
var err error
go func() {
    for e := range errors {
        err = e
    }
}()

for s := range strings {
    fmt.Println(s)
}

return err
```

johnsonj: This ensures we don't jump out as soon as we see the first error while firehose is still trying to send it through. 

The only oddity is if an error occurs at some point in the stream but is recovered from, we will return the last error from this function on normal exit. This might be a case where we want to just log the error channel out to stderr of the nozzle and continue to drain the messages channel till it's closed.

cholick: I think I'd rather go with `FirehoseWithoutReconnect`, so we're being explicit about every error. The 5 retry will mask errors while we're in mvp mode and potentially want to recognize all failure conditions. Bail on first fail until we understand the problems.

src/stackdriver-nozzle/firehose/client.go
nrxus: We don't need to specify an auth token since we are specifying a token refresher. If the auth token is empty, it will retrieve it from the auth token refresher. Yes, we could make the subscription ID configurable. The subscription ID could be added to the configuration.

PR #12, opened 2016-10-03
Adding labels from envelope fields

nrxus:
We also moved the dev functions into runnable go scripts (print firehose to stdout, clear SD metrics, send mock metric to SD every x seconds)


src/stackdriver-nozzle/nozzle/envelope_test.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		labels := envelope.Labels()
+		Expect(labels).To(Equal(map[string]string{
+			"origin": origin,
+			"event_type": eventType.String(),
+			"deployment": deployment,
+			"job": job,
+			"index": index,
+			"ip": ip,
```
johnsonj: shouldn't there be a field here for tags? Or maybe you want to add an (Envelope*) func Payload() that adds the tags there?

src/stackdriver-nozzle/scripts/repro-metrics-rpc-errors.go
```diff
@@ -0,0 +1,70 @@
+package main
```
johnsonj: can you add some documentation to this one? eg, what's the error, why we have this code

src/stackdriver-nozzle/scripts/firehose-to-stdout.go
```diff
@@ -0,0 +1,26 @@
+package main
```
johnsonj: not a hard stance here but i think this one may be appropriate as a part of the main binary. That said we only expect these to get ran by bosh where someone won't use that output so feel free to ignore it.

src/stackdriver-nozzle/config/config.go
```diff
@@ -0,0 +1,42 @@
+package config
```
johnsonj: nice move on breaking these components out

src/stackdriver-nozzle/nozzle/envelope_test.go
nrxus: Not sure - we made a story to figure out what to do with tags here: https://www.pivotaltracker.com/story/show/131571037

src/stackdriver-nozzle/config/config.go
evandbrown: I'd generally prefer flag parsing to happen inside main(). That makes sure anything that needs a flag/config value has to declare it in its method signature, and can't go grabbing initialized package-public vars.

src/stackdriver-nozzle/nozzle/envelope_test.go
johnsonj: cool, thanks :)

src/stackdriver-nozzle/config/config.go
nrxus: I get where you're coming from. We did this to support the scripts, but I think I'm okay having the scripts just work off environment vars anyway.

evandbrown:
Ahh, that makes sense. I'm fine adding a TODO on this one.

PR #13, opened 2016-10-03
Event Filtering

johnsonj:
As an operator, I can configure the types of events that are forwarded to Stackdriver logging.

- Add filter package to only allow specified events through to the output `FirehoseHandler` (eg StdOut or Stackdriver)
- Insert the filter inbetween the input (firehose) and the output (stdout/stackdriver) in the main package
- Add command line options to configure. Defaulting to `LogMessage,Error`

Validation:
- Unit tested the filter package
- Manual validation of the stackdriver-monitoring with --debug option.


src/stackdriver-nozzle/main.go
```diff
@@ -48,21 +54,32 @@ func main() {
...
+	if err != nil || output == nil {
 		panic(err)
 	}
 
+	filteredOutput, err := filter.New(output, strings.Split(*eventsFilter, ","))
+
+	if err != nil {
+		println("Error:", err.Error())
+		filter.DisplayValidEvents()
+		return
```
cholick: https://golang.org/pkg/os/#Exit with non-0 might be a better choice than returning

johnsonj: thanks!

nrxus:
Is it just me, or does the design of the firehose API make the middleware stacking feel backward?

src/stackdriver-nozzle/filter/filter.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	return &f, nil
+}
+
+func (f *filter) HandleEvent(envelope *events.Envelope) error {
+	if f.enabled[envelope.GetEventType()] {
+		return f.dest.HandleEvent(envelope)
+	}
+	return nil
```
nrxus: I might reverse these to keep the happy-path logic less indented, which I think is more idiomatic go.

johnsonj:
@nrxus I thought about that too when I started this. The options I saw to resolve it is to add filtering directly to `firehose.client` or add some sort of callback on the `FirehoseHandler` to grab filters. Those to me added more responsibilities to `firehose` will still requiring a some external filtering logic (parsing event names,..).

I wrote `filter` thinking of it as a decorator on the output but a better way to consider it is probably a pipeline (or middleware as you mentioned). Sure we're filtering the output of `firehose` but this stage is filtering the input of `stackdriver`. 

Totally open to suggestions here though. Perhaps instead of asking for a handler we call it an output (or is there a more idiomatic term for a chain in middleware?)

nrxus:
Yeah, I like the way you went, given the existing firehose API. I (@kejadlen) was just wondering if the code would make more sense (from a reasoning-about-it perspective) if the output took the stack instead of the input? I'm not suggesting that there's anything that actively needs to be done, but wanted to see if it felt kind of strange to anyone else. :)

src/stackdriver-nozzle/filter/filter.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (f *filter) HandleEvent(envelope *events.Envelope) error {
+	if !f.enabled[envelope.GetEventType()] {
+		return nil
+	}
+	return f.dest.HandleEvent(envelope)
+}
+
+func DisplayValidEvents() {
+	println("Valid event choices:")
```
evandbrown: Use fmt.Printf

src/stackdriver-nozzle/main.go
```diff
@@ -48,21 +55,32 @@ func main() {
...
 	}
-	if err != nil {
+	if err != nil || output == nil {
 		panic(err)
 	}
 
+	filteredOutput, err := filter.New(output, strings.Split(*eventsFilter, ","))
+
+	if err != nil {
+		println("Error:", err.Error())
```
evandbrown: Use fmt.Printf

src/stackdriver-nozzle/filter/filter.go
```diff
@@ -0,0 +1,50 @@
...
+type filter struct {
+	dest    firehose.FirehoseHandler
+	enabled map[events.Envelope_EventType]bool
+}
+
+func parseEventName(name string) (events.Envelope_EventType, error) {
+	if eventId, ok := events.Envelope_EventType_value[name]; ok {
+		return events.Envelope_EventType(eventId), nil
+	}
+	return events.Envelope_Error, fmt.Errorf("unknown event name: %s", name)
```
johnsonj: as discussed- can't :(

nrxus: Would it make sense to make your own error struct here and return it with the displayed valid events string so the caller can print that?

johnsonj: Sounds good!

johnsonj:
@nrxus (and @kejadlen?) I totally see what you're saying there. Curious for other feedback, maybe something for cleanup?

johnsonj:
@evandbrown ready for review

nrxus:
LGTM

src/stackdriver-nozzle/main.go
```diff
@@ -42,15 +50,26 @@ var (
...
+	output := nozzle.Nozzle{StackdriverClient: sdClient}
+
+	filteredOutput, err := filter.New(&output, strings.Split(*eventsFilter, ","))
+	if err != nil {
+		if unknownEvent, ok := err.(*filter.UnknownEventName); ok {
+			fmt.Printf("Error: %s, possible choices: %s\n", unknownEvent.Error(), strings.Join(unknownEvent.Choices, ","))
+			os.Exit(-1)
+		} else {
+			panic(err)
+		}
```
nrxus: Might be able to use a [type switch](https://golang.org/doc/effective_go.html#type_switch) here.

johnsonj: I was thinking about that but I couldn't come up with a good generic name to switch on and figured this is straight forward since there's only one expected specialization

nrxus: fair enough, looks okay to us

PR #15, opened 2016-10-05
Add BOSH bits for stackdriver-nozzle

kejadlen:
Not sure if we need logrotate stuff here...

### DIFF
--- a/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
+++ b/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
@@ -18,6 +18,14 @@ chown -R vcap:vcap $LOG_DIR
 exec >>$LOG_DIR/stackdriver-nozzle.stdout.log \
     2>>$LOG_DIR/stackdriver-nozzle.stderr.log
 
+
+export API_ENDPOINT=<%= p('firehose.api_endpoint') %>
+export FIREHOSE_USERNAME=<%= p('firehose.username') %>
+export FIREHOSE_PASSWORD=<%= p('firehose.password') %>
+export SKIP_SSL_VALIDATION=<%= p('firehose.skip_ssl_validation', false) %>
+
+export PROJECT_ID=<%= p('gcp.project_id') %>
+
 chpst -u vcap:vcap /var/vcap/packages/stackdriver-nozzle/bin/stackdriver-nozzle &
 
 echo $! > $PIDFILE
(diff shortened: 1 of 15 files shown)


cholick:
Pretty sure you don't need to deal with logrotate yourself, see
https://bosh.io/docs/job-logs.html#log-rotation

"BOSH log rotates release job logs with the Logrotate log file management utility. Logrotate is configured by the Agent to act on all .log files in the /var/vcap/sys/log/, /var/vcap/sys/log/*/, and /var/vcap/sys/log/*/*/ directories."

jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
```diff
@@ -0,0 +1,46 @@
...
+
+
+export API_ENDPOINT=<%= p('firehose.api_endpoint') %>
+export FIREHOSE_USERNAME=<%= p('firehose.username') %>
+export FIREHOSE_PASSWORD=<%= p('firehose.password') %>
+export SKIP_SSL_VALIDATION=<%= p('firehose.skip_ssl_validation', false) %>
+
+export PROJECT_ID=<%= p('gcp.project_id') %>
+
+chpst -u vcap:vcap /var/vcap/packages/stackdriver-nozzle/bin/stackdriver-nozzle &
```
cholick: Most of the cf managed processes use a `pid_guard` helper to ensure the process doesn't launch more than once. See
* https://github.com/cloudfoundry/cf-release/blob/master/src/common/utils.sh
* https://github.com/concourse/concourse/blob/master/jobs/tsa/templates/tsa_ctl.erb#L39

jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
```diff
@@ -0,0 +1,46 @@
...
+
+echo $! > $PIDFILE
```
cholick: The more common pattern I've seen is to launch in the foreground. See
* https://github.com/concourse/concourse/blob/master/jobs/tsa/templates/tsa_ctl.erb#L47
* https://github.com/cloudfoundry/loggregator/blob/master/jobs/boshhmforwarder/templates/boshhmforwarder_ctl.erb#L24
* https://github.com/cloudfoundry-community/logsearch-for-cloudfoundry/blob/develop/jobs/ingestor_cloudfoundry-firehose/templates/bin/ingestor_cloudfoundry-firehose_ctl#L48

PR #22, opened 2016-10-07
Implement counter events as a metric instead of a log event

nrxus:
Counter events will now be sent to stack driver as metrics instead of logs.

Note that we are using the "total" value of the metric instead of the "delta". 
This might need to be revisited at a later point. I think total is fine for MVP.

### DIFF
--- a/src/stackdriver-nozzle/serializer/serializer.go
+++ b/src/stackdriver-nozzle/serializer/serializer.go
@@ -47,15 +47,16 @@ func (s *cachingClientSerializer) GetLog(e *events.Envelope) *Log {
 }
 
 func (s *cachingClientSerializer) GetMetrics(envelope *events.Envelope) []*Metric {
+	labels := s.buildLabels(envelope)
 	switch envelope.GetEventType() {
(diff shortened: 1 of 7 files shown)


src/stackdriver-nozzle/serializer/serializer.go
```diff
@@ -75,11 +83,8 @@ func (s *cachingClientSerializer) IsLog(envelope *events.Envelope) bool {
 	switch *envelope.EventType {
 	case events.Envelope_HttpStartStop, events.Envelope_LogMessage, events.Envelope_Error:
 		return true
-	case events.Envelope_ValueMetric, events.Envelope_ContainerMetric:
+	case events.Envelope_ValueMetric, events.Envelope_ContainerMetric, events.Envelope_CounterEvent:
 		return false
-	case events.Envelope_CounterEvent:
-		//Not yet implemented as a metric
-		return true
 	default:
```
johnsonj: In that case do we need a matching `IsMetric` to catch the unknowns?

PR #23, opened 2016-10-07
Use the timestamp in  the firehose envelope instead of current time for metrics

nrxus:
We used to use the current time when sending metric events to SD.
We will now use the timestamp in the envelope. Note that we still assume GAUGE type so start and end time of the event is the same.


cholick:
I don't think the units match. Envelope uses nanoseconds and stackdriver uses seconds.

nrxus:
Closing this PR since we are taking another approach.

PR #24, opened 2016-10-07
Use a normal GOPATH

nrxus:
Signed-off-by: Andres Medina <amedina@pivotal.io>


cholick:
The package needs changing too for this change.

```

PR #26, opened 2016-10-10
Refactor of the nozzle + fix of time event

nrxus:
In this PR we are restructuring the code to make the stackdriver client more testable. 
We are also fixing the event timestamp.


src/stackdriver-nozzle/main.go
```diff
@@ -96,17 +95,16 @@ func main() {
 		logger.Fatal("newMetricClient", err)
 	}
 
-	nozzleSerializer := serializer.NewSerializer(cachingClient, logger)
 	metricAdapter := stackdriver.NewMetricAdapter(*projectID, metricClient)
```
johnsonj: sort/group these a bit

src/stackdriver-nozzle/nozzle/handler.go
```diff
@@ -0,0 +1,7 @@
+package nozzle
+
+import "github.com/cloudfoundry/sonde-go/events"
+
+type Handler interface {
```
johnsonj: suggest a different naming strategy, make the interface name specific and the Handle function generic (because the type is already specified in the param type)

johnsonj: Did you consider something like 'Target' or 'Sink' instead of Handler?

PR #29, opened 2016-10-11
Fix stackdriver-nozzle job

evandbrown:
These commits include the dependency for the `pid_guard` function. They also run the `stackdriver-nozzle` process as the default user. Otherwise, the `vcap` user is unable to create the boltdb file required to run.


cholick:
Probably better to `chown` the working and other dirs. See:
https://github.com/cloudfoundry/capi-release/blob/6670022b028d2a032cc5a7369afa806fc8cb8c34/jobs/tps/templates/tps_listener_ctl.erb

PR #37, opened 2016-10-13
Update README

kejadlen:


src/stackdriver-nozzle/README.md
```diff
@@ -1,19 +1,86 @@
+## stackdriver-nozzle
```
johnsonj: Can we get a little blurb here about what it is

src/stackdriver-nozzle/README.md
```diff
@@ -1,19 +1,86 @@
...
+
+`stackdriver-nozzle` is configured through the following environment variables:
+
+#### Firehose
+
+- `FIREHOSE_ENDPOINT` - the CF API endpoint; e.g., `https://api.bosh-lite.com'
+- `FIREHOSE_EVENTS` - comma-separated list of events to pass to Stackdriver;
+  valid events are `LogMessage`, `ValueMetric`, `CounterEvent`, `Error`,
+  `ContainerMetric`, `HttpStart`, `HttpStop`, and `HttpStartStop`
+- `FIREHOSE_USERNAME` - CF username; defaults to `admin`
```
johnsonj: should we put some info on permissions required by this user, here or in its own section?

kejadlen: Yup, that's a good idea... I don't know what they are though. 😅

cholick: openid might not actually be required.

PR #38, opened 2016-10-13
Buffering of Metrics

nrxus:
We only batch based on number of metrics (not time nor size)
We use the same flag as the log batching
Due to API constraints, if we get a time series with the same name/labels as a time series that is already in our buffer, then we send the duplicate by itself since we cannot have multiple points, and cannot have duplicate time serieses (even if the points are different).


src/stackdriver-nozzle/stackdriver/metrics_buffer.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if len(mb.metrics) < mb.size {
+		return
+	}
+
+	mb.postMetrics(mb.metrics)
+	mb.metrics = []Metric{}
+}
+
+func (mb *metricsBuffer) addMetric(newMetric *Metric) {
```
johnsonj: add comment about points limitation here

PR #39, opened 2016-10-13
Update Stackdriver Logging agent and capture syslogs

johnsonj:
- Updated agent version based on https://github.com/GoogleCloudPlatform/google-fluentd/blob/master/plugin_gems.rb
- Added syslog.conf to capture all syslog events (and upload them to Stackdriver Logging)

Confirmed that syslog shows up in the console. Also confirmed that output from the agent itself (stdout/stderr of the job) shows up in Stackdriver.


cholick:
This seems cool, but I'm really unfamiliar with this area of the code. My thought is go ahead and ship it and we'll do some integration testing. One of the stories near the top of our backlog needs this change anyway.

PR #47, opened 2016-10-14
Include running nozzle logs in standard out

cholick:


jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
```diff
@@ -32,8 +32,8 @@ case $1 in
     echo $$ > $PIDFILE
 
     exec chpst -u vcap:vcap /var/vcap/packages/stackdriver-nozzle/bin/stackdriver-nozzle \
-      >>$LOG_DIR/stackdriver-nozzle.stdout.log \
-      2>>$LOG_DIR/stackdriver-nozzle.stderr.log
+      >> >(tee $LOG_DIR/stackdriver-nozzle.stdout.log) \
```
johnsonj: the only oddity to consider here is that the messages from stackdriver-nozzle will get rolled up into the stackdriver-nozzle-ctl log file and will also show up as `vcap.stackdriver-nozzle-ctl` in syslog.

I don't think the syslog category matters but duplicating the logs is probably not necessary.

cholick:
Let's **not** merge this PR, take a look at:
https://www.pivotaltracker.com/story/show/132357265

And investigate a bit first.

jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
kejadlen: We were thinking that we might want a more general purpose solution, but it depends on whether we want to capture logs from other jobs that are logging to files instead of syslog. It seems like something we probably want to do, but I'm not sure.

johnsonj: Ah like route all bosh logs to syslog?

PR #49, opened 2016-10-17
use channels everywhere, clean up main.go

nrxus:


src/stackdriver-nozzle/nozzle/nozzle.go
```diff
@@ -20,17 +21,48 @@ func (e *PostMetricError) Error() string {
...
+			}
+		}
+	}()
+
+	return errs, fhErrs
+}
+
+func (n *Nozzle) Stop() {
+	n.Heartbeater.Stop()
+	n.done <- struct{}{}
```
cholick: Making the done channel bool would read a little better.
`n.done <- true`

nrxus: The way I see the done channel is as receiving a "signal" rather than a value. Sending false to the done channel does not make much sense which is why I am leaning towards an empty struct instead.

PR #53, opened 2016-10-25
Record more info in Heartbeat + Improved Efficiency of Batching

nrxus:
The heartbeat now records:
 * # Metric requests
 * # Metric Timeseries
 * # Logs
 * # Stackdriver Metric Client Errors

Instead of sending a duplicate metric individually, we will now send the current buffer of metrics and start a new one with the duplicate. 
Added a script to get statistic on our heartbeat.

### DIFF
--- a/src/stackdriver-nozzle/heartbeat/heartbeater.go
+++ b/src/stackdriver-nozzle/heartbeat/heartbeater.go
@@ -13,7 +13,7 @@ type Heartbeater interface {
 	Stop()
 }
 
-type heartbeat struct {
+type heartbeater struct {
 	logger  lager.Logger
 	trigger <-chan time.Time
 	counter chan string
@@ -21,10 +21,10 @@ type heartbeat struct {
 	started bool
 }
 
-func NewHeartbeat(logger lager.Logger, trigger <-chan time.Time) Heartbeater {
+func NewHeartbeater(logger lager.Logger, trigger <-chan time.Time) Heartbeater {
 	counter := make(chan string)
 	done := make(chan struct{})
-	return &heartbeat{
+	return &heartbeater{
 		logger:  logger,
 		trigger: trigger,
 		counter: counter,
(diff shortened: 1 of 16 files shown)


src/stackdriver-nozzle/heartbeat/heartbeater.go
```diff
@@ -0,0 +1,73 @@
...
+
+type heartbeater struct {
+	logger  lager.Logger
+	trigger <-chan time.Time
+	counter chan string
+	done    chan struct{}
+	started bool
+}
+
+func NewHeartbeater(logger lager.Logger, trigger <-chan time.Time) Heartbeater {
```
johnsonj: nit: New would be more idiomatic but I know we use this style in other places so feel free to ignore

src/stackdriver-nozzle/heartbeat/heartbeater.go
```diff
@@ -0,0 +1,73 @@
...
+		counters := map[string]uint{}
+		for {
+			select {
+			case <-h.trigger:
+				h.logger.Info(
+					"heartbeater", lager.Data{"counters": counters},
+				)
+				counters = map[string]uint{}
+			case name := <-h.counter:
+				counters[name] += 1
```
johnsonj: very cool

src/stackdriver-nozzle/mocks/heartbeater.go
```diff
@@ -0,0 +1,27 @@
+package mocks
```
johnsonj: this seems like a legit version of the heartbeater, maybe we should just use the real version in tests instead of a mock?

src/stackdriver-nozzle/scripts/metrics_stats.rb
```diff
@@ -0,0 +1,16 @@
+require "json"
```
johnsonj: can you add a comment here that explains what this does/how to run it?

src/stackdriver-nozzle/mocks/heartbeater.go
nrxus: If we did this we would need to pass in a mock of a logger, and our own ticker and test that the logger got the right thing. I do not like this since now the tests for the users of the heartbeater have to know about the logger. Alternatively, we could we change the interface of the heartbeater to trigger a function instead of logging. This is a better but I do not think it is worth doing at this point.

johnsonj:
Thanks didn't consider that dependency

PR #61, opened 2016-12-04
Add metadata.labels to indentify the source of the log line

mrdavidlaing:
This PR extends the behaviour of the google-fluentd configuration such that when it forwards the contents of a log from /var/vcap/sys/log/*/*.log it includes metadata so that the log source can be identified.

Specifically, the following metadata attributes are added

- metadata.labels.deployment
- metadata.labels.job
- metadata.labels.index
- metadata.labels.log_path

This enables searching for specific component logs in StackDriver Logging using a search query like:

```
log="vcap" AND metadata.labels.deployment="cf-c201d9fbf59670f9d707" AND metadata.labels.job="uaa" AND metadata.labels.index="0" 
```

image

### DIFF
--- a/jobs/google-fluentd/templates/google-fluentd.conf
+++ b/jobs/google-fluentd/templates/google-fluentd.conf
@@ -22,4 +22,14 @@
   max_retry_wait 300
   # Disable the limit on the number of retries (retry forever).
   disable_retry_limit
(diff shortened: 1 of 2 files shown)


johnsonj:
Thanks @mrdavidlaing!

PR #65, opened 2017-01-04
Add Stackdriver Errors/Debugger example app

johnsonj:


johnsonj:
Can you give me a sanity check here?

johnsonj:
Thanks! yeah that is the agent that runs the go app. It's just basically vendoring it from [here](https://cloud.google.com/debugger/docs/setting-up-go-on-compute-engine#download_the_agent).

Issue #62, opened 2016-12-14
Throttle TimeSeries requests to Stackdriver metrics API


evandbrown:
The Stackdriver API recommends sending at most 1 TimeSeries value every 30s (a batch request can contain 200 individual TimeSeries values). The nozzle should keep a map of TimeSeries values and put them to the API at a 30s interval.

GarimaSharma:
We are using Opsman 1.11.6, ERT 1.11.7 and stackerdriver nozzle 1.0.3 
We are facing this issue for our setup as well `rpc error: code = 13 desc = stream terminated by RST_STREAM with error code: 2` in our `/var/vcap/sys/log/stackdriver-nozzle/stackdriver-nozzle.stdout.log` logfile.

We noticed that this was fixed in [1.0.2 version](http://docs.pivotal.io/partners/gcp-sdn/release-notes.html#102)
We are still seeing it in 1.0.3 - is there a configuration setting that we have missed to enable this fix?

PR #76, opened 2017-02-08
WIP: Timeseries timer buffer

evandbrown:
> WIP: needs more unit tests

This change allows metrics to be buffered and drained after a duration.
A duplicate metric (i.e., metric name and labels are identical to an
existing buffered metric) overwrites the existing metric it matches.

Addresses #62

### DIFF
--- a/src/stackdriver-nozzle/main.go
+++ b/src/stackdriver-nozzle/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -10,9 +11,9 @@ import (
 
 	"cloud.google.com/go/logging"
 	"github.com/cloudfoundry-community/go-cfclient"
+	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/cloudfoundry"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/config"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/filter"
-	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/cloudfoundry"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/heartbeat"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/nozzle"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/stackdriver"
@@ -23,8 +24,10 @@ import (
 func main() {
 	a := newApp()
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	if a.c.DebugNozzle {
-		defer handleFatalError(a)
+		defer handleFatalError(a, cancel)
 
 		go func() {
 			a.logger.Info("pprof", lager.Data{
@@ -34,7 +37,7 @@ func main() {
 	}
 
 	producer := a.newProducer()
-	consumer := a.newConsumer()
+	consumer := a.newConsumer(ctx)
 
 	errs, fhErrs := consumer.Start(producer)
 	defer consumer.Stop()
@@ -47,12 +50,17 @@ func main() {
 
 	fatalErr := <-fhErrs
 	if fatalErr != nil {
+		// TODO(evanbrown): Give the buffer time to drain before logging fatal
+		cancel()
 		a.logger.Fatal("firehose", fatalErr)
 	}
 }
 
-func handleFatalError(a *app) {
+func handleFatalError(a *app, cancel context.CancelFunc) {
 	if e := recover(); e != nil {
+		// Cancel the context
+		cancel()
+
 		stack := make([]byte, 1<<16)
 		stackSize := runtime.Stack(stack, true)
 		stackTrace := string(stack[:stackSize])
@@ -143,10 +151,10 @@ func (a *app) newProducer() cloudfoundry.Firehose {
 	return producer
 }
 
-func (a *app) newConsumer() *nozzle.Nozzle {
+func (a *app) newConsumer(ctx context.Context) *nozzle.Nozzle {
 	return &nozzle.Nozzle{
 		LogSink:     a.newLogSink(),
-		MetricSink:  a.newMetricSink(),
+		MetricSink:  a.newMetricSink(ctx),
 		Heartbeater: a.heartbeater,
 	}
 }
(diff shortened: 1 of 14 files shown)


src/stackdriver-nozzle/stackdriver/metrics_buffer.go
```diff
@@ -16,12 +16,83 @@
...
+import (
+	"context"
+	"github.com/mitchellh/hashstructure"
+	"reflect"
+	"sync"
+	"time"
+)
 
 type MetricsBuffer interface {
 	PostMetric(*Metric)
```
johnsonj: maybe we want a `Flush`/`IsEmpty` to support [comment](https://github.com/cloudfoundry-community/stackdriver-tools/pull/76/commits/521c7641df5c3c767ca994b860752668a832ed3b#diff-d23e533d7fb5c6f3666b3d9f396c7085R53)

src/stackdriver-nozzle/main.go
```diff
@@ -23,8 +24,10 @@ import (
 func main() {
 	a := newApp()
 
+	ctx, cancel := context.WithCancel(context.Background())
```
johnsonj: yes!

src/stackdriver-nozzle/stackdriver/metric_client.go
```diff
@@ -34,7 +34,9 @@ type MetricClient interface {
 
 func NewMetricClient() (MetricClient, error) {
 	ctx := context.Background()
-	sdMetricClient, err := monitoring.NewMetricClient(ctx, option.WithScopes("https://www.googleapis.com/auth/monitoring.write"), option.WithUserAgent(version.UserAgent()))
+	sdMetricClient, err := monitoring.NewMetricClient(ctx,
```
johnsonj: revert

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
```diff
@@ -78,3 +149,15 @@ func (mb *metricsBuffer) postMetrics(metrics []Metric) {
 		}
 	}()
 }
+
+type metricsMap map[string]*Metric
+
+func (mm metricsMap) Slice() []Metric {
+	m := make([]Metric, len(mm))
+	var i int
+	for _, v := range mm {
```
johnsonj: can you use the `i` that range gives you?

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
```diff
@@ -16,12 +16,83 @@
...
+	"reflect"
+	"sync"
+	"time"
+)
 
 type MetricsBuffer interface {
 	PostMetric(*Metric)
 }
 
+type timerMetricsBuffer struct {
```
johnsonj: CullingMetricsBuffer? SamplingMetricsBuffer?

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
```diff
@@ -16,12 +16,83 @@
...
+
+func (mb *timerMetricsBuffer) PostMetric(metric *Metric) {
+	mb.addMetric(metric)
+}
+
+func (mb *timerMetricsBuffer) addMetric(newMetric *Metric) {
+	k, _ := hashstructure.Hash(newMetric.Labels, nil)
+	mb.metricsMu.Lock()
+	defer mb.metricsMu.Unlock()
+	mb.metrics[newMetric.Name+string(k)] = newMetric
```
johnsonj: this is the magic line right? where we overwrite an existing metric reading if it's in the buffer and we haven't flushed yet. 

pls comment here :)

evandbrown: Yep! Worth noting I've confirmed that `hashstructure` consistently hashes labels, even though that's a `map[string]string` with no guaranteed order.

This is a naive take. We could make `metrics` a `map[string][]*Metric`, appending dupes and allowing something downstream to decide how to handle the dupes (average them, take earliest, latest, etc). I haven't thought much about where we'd do that, but probably here or the adapter.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: `mm` is a map so the first value in the tuple is the map key :(

johnsonj: doh! carry on

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: Down yonder in [`start`](https://github.com/cloudfoundry-community/stackdriver-tools/blob/521c7641df5c3c767ca994b860752668a832ed3b/src/stackdriver-nozzle/stackdriver/metrics_buffer.go#L82) we're receiving from the context's `Done` channel and flushing metrics there. I think I've got a TODO in `main.go` to give that a few seconds from the time we cancel to complete.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: `periodicMetricsBuffer`?

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
johnsonj: Suggest explicitly checking over a random wait period. Perhaps bock on buffer.HasMetrics() with a timeout. Just gives some more certainty and prevents any unneeded sleeping.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
johnsonj: Depends on if this is the class that does the sampling (or averaging, ..) or not. To me that seems like a big deal and should be part of the name.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
johnsonj: My first take on this is that we don't have the ability to make smarts decision about combining metrics. Metrics may be dropped in transit and they will for sure be sprayed across multiple nozzle instances. Some metrics are already averages, or cumulative totals, or deltas since the last, .. As a result any sort of calculation we do is just adding errors together. So the simpler our sampling is, the better. 

Perhaps I'm missing a smarter solution.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: Ahh, gotcha! I missed the link to the comment. Makes sense.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: No sampling. I like it.

src/stackdriver-nozzle/stackdriver/metrics_buffer.go
evandbrown: Yeah, so what are we actually doing? Seems like we're buffering metrics, culling them, and periodically flushing them. `autoCulledMetricsBuffer`?

PR #83, opened 2017-02-23
Instrument nozzle stats

evandbrown:
This change extends the heartbeater to write events to Stackdriver metrics. It writes delta metrics at the heartbeat interval to provide timeseries data about heartbeat event frequency. Heartbeat handling is abstracted to support the logger (without affecting the existing heartbeat API) and metrics outputs. 

Functionality from the `stackdriver` package was duplicated in the `heartbeat` package because reusing the former from the latter created a circular dependency.


src/stackdriver-nozzle/heartbeat/metric_handler.go
```diff
@@ -0,0 +1,74 @@
...
+	defer h.counterMu.Unlock()
+
+	now := time.Now()
+	metrics := []stackdriver.Metric{}
+	for k, v := range h.counter {
+		metrics = append(metrics, stackdriver.Metric{
+			Name:  "nozzle-heartbeat/" + k,
+			Value: float64(v),
+			Labels: map[string]string{
+				"instance": "TODO",
```
evandbrown: I still need to todo this

evandbrown:
I todid all of my todos; ready for your LGTM, @johnsonj

src/stackdriver-nozzle/config/config.go
```diff
@@ -96,6 +101,24 @@ func (c *Config) ensureProjectID() error {
 	return nil
 }
 
+// If running on GCE, this will set the nozzle's ID to the VM ID.
+func (c *Config) setNozzleId() {
+	if metadata.OnGCE() {
```
johnsonj: i like this!

Issue #99, opened 2017-08-17
Support exporting metrics to Stackdriver Logging


johnsonj:
Today we [split](https://github.com/cloudfoundry-community/stackdriver-tools/blob/master/src/stackdriver-nozzle/nozzle/nozzle.go#L79) events from the Loggregator into Stackdriver Logging/Monitoring. It is desirable to have the metric data available in Stackdriver Logging so users can use exports for further analysis. An example would be exporting to BigQuery to perform custom calculations on metrics not available in Stackdriver Monitoring. 

Design Considerations:
- Should it be a 'All or Nothing' to log to both Stackdriver Logging and Monitoring or should each endpoint have it's own list of events? 
- Should we need to perform the same culling for metrics destined for Stackdriver Logging?

mrdavidlaing:
> Should it be a 'All or Nothing' to log to both Stackdriver Logging and Monitoring or should each endpoint have it's own list of events?

We were certainly surprised that when adding `CounterEvent,ValueMetric,ContainerMetric` events to [`firehose.events:`](https://github.com/cloudfoundry-community/stackdriver-tools/blob/master/jobs/stackdriver-nozzle/spec#L21) sent them to a different place ( Stackdriver Monitoring rather than Stackdriver Logging ).

We'd like to propose having 2 separate config settings ( in both the BOSH release and the Tile )

```
  firehose.events_to_stackdriver_logging:
    description: Whitelisted event types to send to Stackdriver Logging (comma separated). Valid values include LogMessage,Error,HttpStartStop,CounterEvent,ValueMetric,ContainerMetric
    default: LogMessage,Error
  firehose.events_to_stackdriver_monitoring:
    description: Whitelisted event types to send to Stackdriver Monitoring (comma separated). Valid values include CounterEvent,ValueMetric,ContainerMetric
    default: CounterEvent,ValueMetric,ContainerMetric
```

For the Pivotal Tracker on GCP use case; we'd probably send everything to Stackdriver Logging -> because we want the ability to send some of the data into BigQuery, and `CounterEvent,ValueMetric` to Stackdriver Monitoring

mrdavidlaing:
@johnsonj Thanks for acting on this issue so quickly.

Will this change make its way into v1.0.6 of the [GCP Stackdriver Nozzle for PCF tile](https://network.pivotal.io/products/gcp-stackdriver-nozzle) ?

johnsonj:
[...]

PR #117, opened 2017-09-19
Send Metrics to Stackdriver Logging + Perf improvements to handle it

johnsonj:
Background:
Allow users to send metrics to Stackdriver Logging.

This was attempted in #99 but the solution would crash after a short time and hang. The hang was fixed in #115. This fixes the performance by taking a new approach.

Changes:
- Introduce the `MetricRouter`: This sits near the end of the metric pipeline, after the `AutoCullingMetricBuffer` but before the final `MetricAdapater` that writes to Stackdriver. This router decides where the finalized Metric (hydrated with labels, units, ..) gets sent to (Stackdriver Logging and/or Monitoring). See: b289b73
- Make the `AutoCullingMetricBuffer.flush()` not block metric arrival. Previously the buffer could not receive any new messages while it was being flushed. See: c352728

Refactoring:
- `stackdriver` package now contains only code for writing to Stackdriver. The message (Log/Metric) types are extracted out. The `AutoCullingMetricBuffer` is extracted out. 
- Explicitly make `MetricsBuffer` inherit `MetricAdapater` and remove knowledge of the buffer from the nozzle.


src/stackdriver-nozzle/nozzle/metric_sink.go
```diff
@@ -47,38 +49,41 @@ func (ms *metricSink) Receive(envelope *events.Envelope) error {
...
 
-	var metrics []stackdriver.Metric
+	var metrics []messages.Metric
 	switch envelope.GetEventType() {
 	case events.Envelope_ValueMetric:
 		valueMetric := envelope.GetValueMetric()
-		metrics = []stackdriver.Metric{{
+		metrics = []messages.Metric{{
 			Name:      valueMetric.GetName(),
+			Type:      envelope.GetEventType(),
```
evandbrown: I probably missed this somewhere else in the CL, but what does the new `Type` field get us?

src/stackdriver-nozzle/nozzle/parse_events.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+
+	"github.com/cloudfoundry/sonde-go/events"
+)
+
+func ParseEvents(names []string) ([]events.Envelope_EventType, error) {
+	events := []events.Envelope_EventType{}
+
+	for _, name := range names {
+		if name == "" {
```
evandbrown: This would allow the config to contains something like `foo,,bar`. Do we want that, or should it error instead of continuing?

src/stackdriver-nozzle/nozzle/parse_events.go
```diff
@@ -0,0 +1,47 @@
...
+		return events.Envelope_EventType(eventId), nil
+	}
+	return events.Envelope_Error, &invalidEvent{name: name}
+}
+
+type invalidEvent struct {
+	name string
+}
+
+func (ie *invalidEvent) Error() string {
```
evandbrown: nice

src/stackdriver-nozzle/nozzle/metric_sink.go
johnsonj: I needed it in in the [router](https://github.com/cloudfoundry-community/stackdriver-tools/pull/117/commits/b289b73e42b1284685f838e509f7db6dec42fa7d#diff-b5bfcc892bab6eeaa9fcaf2add9391bbR36) to determine where to send it. I'm not crazy about it because it adds a specific type from loggregator into the internal representation. If you have other ideas I'm all ears.

src/stackdriver-nozzle/nozzle/parse_events.go
johnsonj: That's a good point. I don't think we want to explicitly allow that. The alternate here is to special case empty string [here](https://github.com/cloudfoundry-community/stackdriver-tools/pull/117/files#diff-f484551b1a038698c9d7f1840e368a1cR90) (or move the parsing into the config object).

The reason I added this is because: `strings.Split("", ",") => []string{""}`

johnsonj:
Thank you @evandbrown ! Nothing too controversial so I'm going to merge.

PR #134, opened 2017-10-31
Include metric names while deduplicating values in the buffer.

knyar:
This will make sure that each set of metrics is deduplicated ("culled")
independently even if they share the same label values.


src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("autoCulledMetricsBuffer", func() {
 		subject.PostMetricEvents([]*messages.MetricEvent{
 			{
 				Labels:  map[string]string{"Name": "a"},
-				Metrics: []*messages.Metric{{Name: "a", Value: 1}},
+				Metrics: []*messages.Metric{{Name: "a", Value: 1}, {Name: "b", Value: 0}},
```
johnsonj: great demonstration of why this is needed

PR #139, opened 2017-11-03
Prepend event origin to all generated metric names.

fluffle:
[Pivotal's docs][1] use the origin to differentiate between metrics that
have the same name, explicitly prepending it to the metric name. The
DataDog nozzle does the same thing. Consistency with these is valuable,
though this does introduce a change in metric naming that is not
backwards-compatible.

Example name changes, first the "now matches the docs, but whatever":

*   AuctioneerTaskAuctionsFailed → auctioneer.AuctioneerTaskAuctionsFailed
*   LRPsExtra → bbs.LRPsExtra

Then the "wow, disambiguation is helpful!":

*   RequestLatency → bbs.RequestLatency
*   latency → gorouter.latency
*   Latency → etcd.Latency
*   requests.completed → cc.requests.completed
*   total_requests → gorouter.total_requests
*   ReceivingRequestRate → etcd.ReceivingRequestRate

Then the "maybe we don't want to do this, ugh":

*   numGoRoutines →
    *   tps_listener.numGoRoutines
    *   tcp_emitter.numGoRoutines
    *   ... etc

The last one is different because numGoRoutines means the same thing
across multiple origins. Using a label to differentiate instead here
is probably the right thing to do, but conditionally appending the
origin requires more complexity to enumerate where that condition holds.

[1]: https://docs.pivotal.io/pivotalcf/1-12/monitoring/kpi.html

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/139)

### DIFF
--- a/src/stackdriver-nozzle/nozzle/metric_sink.go
+++ b/src/stackdriver-nozzle/nozzle/metric_sink.go
@@ -43,6 +43,9 @@ type metricSink struct {
 func (ms *metricSink) Receive(envelope *events.Envelope) error {
 	labels := ms.labelMaker.Build(envelope)
 
+	// Origin is a required field so this is fine.
+	originPrefix := envelope.GetOrigin() + "."
+
 	timestamp := time.Duration(envelope.GetTimestamp())
 	eventTime := time.Unix(
 		int64(timestamp/time.Second),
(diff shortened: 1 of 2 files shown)


johnsonj:
LGTM

Great. The disambiguation is critical. In the numGoRoutines example I don't even think it's very useful to look at that metric across the multiple pieces of software. tps_listener might be written in a way that it's expected to use tons of goroutines and tcp_emitter isn't suppose to use any. 

We shouldn't get the nozzle too far into the business of massaging metrics.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/139#-:-Ky2_w8IpQ70kKWe7s8i:b-yaz668)*

PR #140, opened 2017-11-03
Log stackdriver request proto along with errors.

fluffle:
Errors returned by Stackdriver reference the request proto, so it is helpful to have this logged alongside the errors.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/140)

### DIFF
--- a/src/stackdriver-nozzle/app/builder.go
+++ b/src/stackdriver-nozzle/app/builder.go
@@ -27,6 +27,10 @@ type App struct {
 	bufferEmpty func() bool
 }
 
+type HasData interface {
+	Data() lager.Data
+}
+
 func New(c *config.Config, logger lager.Logger) *App {
 	logger.Info("version", lager.Data{"name": version.Name, "release": version.Release(), "user_agent": version.UserAgent()})
 	logger.Info("arguments", c.ToData())
@@ -157,7 +161,11 @@ func (a *App) newMetricSink(ctx context.Context, metricAdapter stackdriver.Metri
 	a.bufferEmpty = metricBuffer.IsEmpty
 	go func() {
 		for err := range errs {
-			a.logger.Error("metricsBuffer", err)
+			if errData, ok := err.(HasData); ok {
+				a.logger.Error("metricsBuffer", err, errData.Data())
+			} else {
+				a.logger.Error("metricsBuffer", err)
+			}
 		}
 	}()
 
--- a/src/stackdriver-nozzle/stackdriver/metric_adapter.go
+++ b/src/stackdriver-nozzle/stackdriver/metric_adapter.go
@@ -24,15 +24,12 @@ import (
 	"time"
 
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/messages"
+	"github.com/cloudfoundry/lager"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	labelpb "google.golang.org/genproto/googleapis/api/label"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 )
-
-type MetricAdapter interface {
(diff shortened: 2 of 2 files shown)


johnsonj:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/stackdriver-nozzle/app/builder.go, line 30 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/140#-Ky2gDY3aBOLK1SNLM3V:-Ky2gDY3aBOLK1SNLM3W:b-mw2yuw) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/2d2a3420a0657ab23733f76830387d38f3d7a7d6/src/stackdriver-nozzle/app/builder.go#L30)):*
> ```Go
> }
> 
> type HasData interface {
> ```

This feels too magical to me. Would it make more sense to expose `PostMetricErr` in  the `stackdriver` package? 

The app package knows everyone so it wouldn't be a new dependency.

---

*[src/stackdriver-nozzle/stackdriver/metric_adapter.go, line 170 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/140#-Ky2gieDN_g2_cEwehZz:-Ky2gieDN_g2_cEweh_-:b-rgj55k) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/2d2a3420a0657ab23733f76830387d38f3d7a7d6/src/stackdriver-nozzle/stackdriver/metric_adapter.go#L170)):*
> ```Go
> 	compositeErr.request = request
> 	compositeErr.postErr = ma.client.Post(request)
> 	return compositeErr.processErrors()
> ```

this makes more sense to me, good move

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/140)*

fluffle:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/stackdriver-nozzle/app/builder.go, line 30 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/140#-Ky2gDY3aBOLK1SNLM3V:-KyFgGA3nmx5V6N22WEd:bkvk204) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/2d2a3420a0657ab23733f76830387d38f3d7a7d6/src/stackdriver-nozzle/app/builder.go#L30)):*

This feels too magical to me. Would it make more sense to expose `PostMetricErr` in  the `stackdriver` package? 

The app package knows everyone so it wouldn't be a new dependency.
</blockquote></details>

I'll change it if you want, but I hope I can convince you otherwise :-)

I realise this does seem a little magical but it's one of the really useful outcomes of types automatically fulfilling interfaces. It makes if much easier for the interface to be a contract specified by the *user* rather than one specified by the implementer. Here, it allows the app package to say "if you want to write structured data, expose a HasData method that returns a lager.Data", and then as long as custom error types created here fulfil both error and HasData, they can write structured logs with no extra code. 

If we choose to expose PostMetricErr, then we have to choose to expose all custom error types from all packages that might bubble up to the app package via error channels, and we have to handle all of them individually with a type switch. That's not as scalable as this approach, if more custom error types appear in the future.

Meta-comment: there's a mix of "push a logger down into the package and log errors when you receive them" and "bubble errors up via a channel and log them near main" in the nozzle currently. I chose this approach because it was the easiest way to move more information through the pre-existing error channel, but IMO standardising on the former strategy is the best thing to do. That way, useful backtraces can be logged from "fatal" errors, and there's no need to jump through these magical interface hoops because packages can just write structured data to the log directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/140)*

fluffle:
Actually, having seen your PR #143 I don't think this is necessary, because you're pushing the logger down into the metrics adapter so it's possible to simply log the request proto there when necessary.

PR #141, opened 2017-11-03
Cull metrics based on event timestamp.

fluffle:
The firehose does not guarantee in-order message delivery,
so it is possible for the nozzle to receive older metrics with
the same Hash() after newer ones.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/141)


johnsonj:
LGTM. If you think it's better to keep the newest stuff then let's do it. 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/141#-:-Ky2edBEiYHQ1SHFfVYC:bi118vd)*

PR #136, opened 2017-10-31
Reduce number and increase consistency of metric/log labels.

fluffle:
Stackdriver (SD) imposes a limit of 10 labels per custom metric. Simply copying event metadata into SD labels results in some metrics hitting that limit.
    
To address this, drop eventType and the org, space, and application UUIDs from metric labels completely. Create an "applicationPath" label that concatenates together the org, space and application name.
    
Keep the VM and application index labels separate so that higher level aggregations of the same metric data can be created.
    
Distinguish between the labels applied to log messages and the labels applied to metrics, because distinguishing logs by origin and event type is useful and there doesn't appear to be a similar label limit for SD logging.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/136)

PR #144, opened 2017-11-06
 Config options for metric prefix and director name.

fluffle:
The nozzle currently places all the metrics it derives from firehose
data in the root of Stackdriver's custom metric namespace. This means
there's a relatively large chance that these names could collide with
others that were not created by the nozzle. To mitigate this, a config
option "metric_path_prefix" (that defaults to "firehose") is added.
After this commit, Stackdriver metric names will be of the form:

    custom.googleapis.com/PREFIX/origin.Name
    custom.googleapis.com/firehose/gorouter.total_requests

Running multiple PCF instances in the same GCP project will result in
metrics from all instances being confused with each other. Since the
nozzle runs on BOSH, another config option "bosh_director_name" (that
defaults to "cf") has been added. This sets the value of a static
"director" label added to every metric exported by the nozzle. Setting
this to different values for each PCF instance in a project (e.g. the
 GCP region, if running one PCF instance per region) allows PCF metrics
to be distinguished from one another.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/144)

### DIFF
--- a/jobs/stackdriver-nozzle/spec
+++ b/jobs/stackdriver-nozzle/spec
@@ -51,3 +51,11 @@ properties:
   nozzle.resolve_app_metadata:
     description: Enable resolution of app metadata from appGuid
     default: true
+
+  nozzle.metric_path_prefix
+    description: Prefix added to all metric names being sent to Stackdriver, e.g. 'custom/PREFIX/gorouter.total_requests'. May contain slashes.
+    default: firehose
+
+  nozzle.bosh_director_name
+    description: Name added as the 'director' label to all time series being sent to Stackdriver, useful for differentiating between multiple PCF instances in a project.
+    default: cf
--- a/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
+++ b/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
@@ -34,6 +34,8 @@ case $1 in
     export RESOLVE_APP_METADATA=<%= p('nozzle.resolve_app_metadata', 'true') %>
     export METRICS_BUFFER_DURATION=<%= p('nozzle.metrics_buffer_duration', '30') %>
     export METRICS_BATCH_SIZE=<%= p('nozzle.metrics_batch_size', '200') %>
+    export METRIC_PATH_PREFIX=<%= p('nozzle.metric_path_prefix', 'firehose') %>
+    export BOSH_DIRECTOR_NAME=<%= p('nozzle.bosh_director_name', 'cf') %>
 
     <% if_p('gcp.project_id') do |prop| %>
     export GCP_PROJECT_ID=<%= prop %>
--- a/src/stackdriver-nozzle/README.md
+++ b/src/stackdriver-nozzle/README.md
@@ -39,6 +39,10 @@ go get github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzl
 
 #### Nozzle
 
+- `BOSH_DIRECTOR_NAME` - sets the value of the "director" label added to every
+  metric / log exported to Stackdriver; defaults to "cf". This is useful for
+  differentiating between multiple cloud foundry / BOSH instances in the same
+  GCP / Stackdriver project.
 - `HEARTBEAT_RATE` - how often `stackdriver-nozzle` reports stats to stdout;
   defaults to 30 seconds
 - `LOGGING_BATCH_COUNT` - how many logs to batch into a single report to
@@ -49,6 +53,10 @@ go get github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzl
   buffer; defaults to 30
 - `METRICS_BATCH_SIZE` - batch size for metric time series being sent to
   Stackdriver; defaults to 200
+- `METRIC_PATH_PREFIX` - sets a prefix for all custom metrics exported to
+  Stackdriver, e.g. custom.googleapis.com/PREFIX/gorouter.total_requests;
+  defaults to "firehose". May contain slashes. Useful to "namespace"
+  cloud foundry metrics from others in the same Stackdriver project.
 - `RESOLVE_APP_METADATA` - whether to hydrate app UUIDs into org name, org
   UUID, space name, space UUID, and app name; defaults to `true`
 - `SUBSCRIPTION_ID` - what subscription ID to use for connecting to the
--- a/src/stackdriver-nozzle/app/builder.go
+++ b/src/stackdriver-nozzle/app/builder.go
@@ -70,7 +70,7 @@ func New(c *config.Config, logger lager.Logger) *App {
 	} else {
 		appInfoRepository = cloudfoundry.NullAppInfoRepository()
 	}
-	labelMaker := nozzle.NewLabelMaker(appInfoRepository)
+	labelMaker := nozzle.NewLabelMaker(appInfoRepository, c.BoshDirectorName)
 
 	return &App{
 		logger:      logger,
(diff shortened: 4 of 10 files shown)


fluffle:
I'll rebase this onto develop after PR #136 is in, which ought to make it easier to review. I wanted to base this PR off that one because it made logical sense to do so, but I'm not github-savvy enough to know how to exclude the first commit from the second PR, if such a thing is even possible.

knyar:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/config/config.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-KyGR5WZiss4RJvWHxTz:-KyGR5WZiss4RJvWHxU-:b-s0d7wh) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3d58d4a8569e36c4553fa8681e2658c17d4ab5e2/src/stackdriver-nozzle/config/config.go#L69)):*
> ```Go
> 	MetricsBufferSize     int    `envconfig:"metrics_buffer_size" default:"200"`
> 	MetricPathPrefix      string `envconfig:"metric_path_prefix" default:"firehose"`
> 	BoshDirectorName      string `envconfig:"bosh_director_name" default:"cf"`
> ```

What I actually had in mind is slightly more flexible: instead of having a single hard-coded label with user-supplied value, what if we could just accept an arbitrary number of key/value pairs that are passed as labels for all metrics? This will allow users to attach other instance-specific metadata to their metrics. There is obviously a risk that providing too many labels will exceed 10 label limit, but this will be clearly visible in the logs (and we could document this as a caveat). What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144)*

fluffle:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/config/config.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-KyGR5WZiss4RJvWHxTz:-KyGXoNIbYW3FjPlexce:b-tk5cct) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3d58d4a8569e36c4553fa8681e2658c17d4ab5e2/src/stackdriver-nozzle/config/config.go#L69)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

What I actually had in mind is slightly more flexible: instead of having a single hard-coded label with user-supplied value, what if we could just accept an arbitrary number of key/value pairs that are passed as labels for all metrics? This will allow users to attach other instance-specific metadata to their metrics. There is obviously a risk that providing too many labels will exceed 10 label limit, but this will be clearly visible in the logs (and we could document this as a caveat). What do you think?
</blockquote></details>

I did originally do that, but stepped back, because: why?

I don't see any value in attaching >1 label to all metrics from a PCF instance, because it should be possible to derive any other values you might want to for that instance based on the one label. (I guess SD doesn't quite have JoinWithLiteralTable yet, but that's what I'm thinking of.)

This also enforces the label name to be "director" which is what the PCF folks want to standardise on as the label to differentiate between PCF instances, according to David Laing. Only allowing users to change the value enforces naming consistency.

Then there's the risk of exceeding the 10 label limit. There's some headroom there but I am not a fan of leaving guns footwards like that...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144)*

johnsonj:
@fluffle - I rebased your changed on develop and force pushed. I don't know if that's proper etiquette or not

johnsonj:
Can you also update: `tile.yml.erb`, `jobs/stackdriver-nozzle/spec` with the new fields? 

---

Reviewed 7 of 10 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-:-KyHlXyZj4Yeek0XT8wN:bqpv8ls)*

johnsonj:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-KyHmwexO1hk1QGDTFsk:-KyHmwexO1hk1QGDTFsl:bxkm0pj) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/21f1847e440febac1c681a58754f14e48061bfd5/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb#L38)):*
> ```erb
>     export METRICS_BUFFER_SIZE=<%= p('nozzle.metrics_buffer_size', '200') %>
>     export METRIC_PATH_PREFIX=<%= p('nozzle.metric_path_prefix', 'firehose') %>
>     export BOSH_DIRECTOR_NAME=<%= p('nozzle.bosh_director_name', 'cf') %>
> ```

Can you also update: tile.yml.erb, jobs/stackdriver-nozzle/spec with the new fields?

(reposting as comment so it can be resolved) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144)*

fluffle:
I had to force-push updates anyway. I think @knyar's comments may have been on changes that were in the commit from #136, not sure. Hope I've got everything now!

---

Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-KyHmwexO1hk1QGDTFsk:-KyKiPG7Tqzj8jv-INIO:bati3my) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/21f1847e440febac1c681a58754f14e48061bfd5/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb#L38)):*

Can you also update: tile.yml.erb, jobs/stackdriver-nozzle/spec with the new fields?

(reposting as comment so it can be resolved) 
</blockquote></details>

Aha, thank you. I should have done a recursive grep to see if there was more wiring to be connected up :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-:-KyKiXJwgevroy6VSN2Y:b15qu7n)*

johnsonj:
:lgtm:

---

Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-KyHmwexO1hk1QGDTFsk:-KyMu3lgkeRGBb-GTAyP:b-ucfknh) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/21f1847e440febac1c681a58754f14e48061bfd5/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb#L38)):*

Aha, thank you. I should have done a recursive grep to see if there was more wiring to be connected up :-)
</blockquote></details>

Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144#-:-KyMu0Y5L80z2RIeisb5:bnfp4nl)*

johnsonj:
Reviewed 8 of 8 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/144)*

Issue #142, opened 2017-11-03
Use native Stackdriver Monitoring Protos


johnsonj:
see discussion in #137. Use native protos at the point of [translation](https://github.com/cloudfoundry-community/stackdriver-tools/blob/develop/src/stackdriver-nozzle/nozzle/metric_sink.go#L43) instead of ushering them into an intermediary format.

fluffle:
I started looking at this in more detail, because of @knyar 's comments on #156. The separation of Metric from MetricDescriptor in the [API protos](https://godoc.org/google.golang.org/genproto/googleapis/api/metric) complicates matters somewhat, because we must derive the descriptors from firehose data.

I think (based on an hour or so of poking around, so don't take this as gospel) that the correct route forward is to respect this separation. Handwaving proto-design-doc discussion follows!

We should create a stackdriver.MetricRegistry (or DescriptorRegistry, or whatever) to encapsulate the creation and management of MetricDescriptors. This:

*   Removes this responsibility from the stackdriver.MetricAdapter, so it is solely responsible for batching and posting the timeseries data to Stackdriver. 
*   Makes the error handling in MetricAdapter cleaner still, because we no longer have two types of error to be surfacing (descriptor creation and timeseries creation).

The Metric Type is the shared key tying a Metric to its Descriptor in the API; we can use that in the MetricRegistry for deduping and caching as is done in the MetricAdapter today. I don't think anything more than:

```Go
type MetricRegistry interface {
    // Checks descriptor.Type against internal cache
    // Returns error if cached and descriptors don't match
    // Creates if not cached
    // Adds to cache if successful
    // Returns error on failure
    // May not even need created bool, but probably useful for testing.
    Register(descriptor *metricpb.MetricDescriptor) (created bool, err error)
}
```

is needed, but I have not done enough due diligence to be 100% sure. An alternative API might create separate Register functions per MetricKind or maybe MetricKind+ValueType and take proto elements as args instead:

```Go
type MetricRegistry interface {
    RegisterGauge(metricType string, valueType MetricDescriptor_ValueType, labels []*LabelDescriptor,  otherStuff ...string) (created bool, err error)
    // ... or ...
    RegisterInt64Gauge(metricType string, labels []*LabelDescriptor, otherStuff ...string) (created bool, err error)
    // ... etc, etc.
}

[...]

PR #147, opened 2017-11-07
stackdriver-nozzle/tile: debugging options, logging configuration

johnsonj:
- Expose `NOZZLE_DEBUG` setting to allow for pprof access on the hosted
  nozzle (http://<VM>:6060/debug/pprof/). This also enables Stackdriver Error reporting for crashes. Fixes #114
- Expose Logging buffer duration/size settings and update default to a more reasonable value
- Add Labels for recently added options in the tile for better presentation. At some point we'll want to separate these settings into logical pages now that the options are getting long.

/cc @knyar @fluffle

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/147)

PR #156, opened 2017-11-15
Merge Metric and MetricEvent (#142)

knyar:
`MetricEvent` seems like an artifact of Firehose data model (`Envelope` being separate from the metric itself) which is quite unusual and difficult to reason about. I would like to propose merging `Metric` and `MetricEvent` structs.

I've realized that `router.PostMetrics` also sends metric events to Stackdriver logging, and this changes the payload. Please let me know if you think it's a concern.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/156)

### DIFF
--- a/src/stackdriver-nozzle/messages/metric.go
+++ b/src/stackdriver-nozzle/messages/metric.go
@@ -8,31 +8,24 @@ import (
 	"github.com/cloudfoundry/sonde-go/events"
 )
 
+// Metric represents one of the metrics contained in an events.Envelope.
 type Metric struct {
 	Name      string
+	Labels    map[string]string `json:"-"`
 	Value     float64
 	EventTime time.Time
-	Unit      string // TODO Should this be "1" if it's empty?
+	Unit      string                    // TODO Should this be "1" if it's empty?
+	Type      events.Envelope_EventType `json:"-"`
 }
 
-// MetricEvent represents the translation of an events.Envelope into a set
-// of Metrics
-type MetricEvent struct {
-	Labels  map[string]string `json:"-"`
-	Metrics []*Metric
-	Type    events.Envelope_EventType `json:"-"`
-}
-
-func (m *MetricEvent) Hash() string {
+func (m *Metric) Hash() string {
 	var b bytes.Buffer
 
 	// Extract keys to a slice and sort it
-	numKeys := len(m.Metrics) + len(m.Labels)
+	numKeys := len(m.Labels) + 1
 	keys := make([]string, numKeys, numKeys)
-	for _, m := range m.Metrics {
-		keys = append(keys, m.Name)
-	}
-	for k, _ := range m.Labels {
+	keys = append(keys, m.Name)
+	for k := range m.Labels {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
--- a/src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go
+++ b/src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go
@@ -43,7 +43,7 @@ type autoCulledMetricsBuffer struct {
 	logger  lager.Logger
 
 	metricsMu sync.Mutex // Guard metrics
-	metrics   map[string]*messages.MetricEvent
+	metrics   map[string]*messages.Metric
 }
 
 // NewAutoCulledMetricsBuffer provides a MetricsBuffer that will cull like metrics over the defined frequency.
@@ -51,7 +51,7 @@ type autoCulledMetricsBuffer struct {
 func NewAutoCulledMetricsBuffer(ctx context.Context, logger lager.Logger, frequency time.Duration, adapter stackdriver.MetricAdapter) MetricsBuffer {
 	mb := &autoCulledMetricsBuffer{
 		adapter: adapter,
-		metrics: make(map[string]*messages.MetricEvent),
+		metrics: make(map[string]*messages.Metric),
 		ctx:     ctx,
 		logger:  logger,
 		ticker:  time.NewTicker(frequency),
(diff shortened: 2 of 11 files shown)


knyar:
Taking another look, I've realized that Stackdriver logging is exactly the reason why this has been done on purpose (https://github.com/cloudfoundry-community/stackdriver-tools/pull/120/commits/4790b0c25929455cd59207fc3f8caeadcebd9236), so feel free to reject this. In any case, it might be better to just switch to using Stackdriver protos in a single change rather than merging Metric and MetricEvent first (which was still useful learning exercise).

However, I've realized that I don't understand the value of passing metric events through the metric buffer (which drops duplicate metrics within a given buffer interval) before sending them to the log adapter. If someone is interested in using the Nozzle to pass all events (even metric events) to SD logging, won't it be confusing that the nozzle drops some of them? Should metric events that need to be logged just be routed through the logSink -> logAdapter directly, instead of going through the metric buffer like they do now? (pic below shows current state)

sd nozzle

johnsonj:
Excellent diagram @fluffle. This captures it very well. 

When I first implemented 'Metrics To Stackdriver Logging' I did it the most straightforward way: Just send the firehose events through the logging pipeline. What I found was the volume of metrics (esp. ContainerMetric) is incredibly high. This causes the Logging stream to be unnecessarily all metric related and caused the nozzle to be less reliable. Sending it through the buffer solved this problem in the same way we solved it for Stackdriver Monitoring. 

I'm not opposed to back pedaling on this as it may not be a performance issue now that Logging has more appropriate batch settings (#147). Do you think there is value in reporting them as often as the are from Loggregator? 

The other concern I have is if we get consistent nozzling and begin to report cumulative values instead of gauge. This logic will keep track of start time and modify what we report to Stackdriver Monitoring. I believe we will want the same modified output to Stackdriver Logging and I have so far imagined it happening in the buffer (or a buffer-like part of the `metrics_pipeline`)

fluffle:
My perspective is the same as @knyar 's: the Nozzle deduping (and thus silently throwing away) firehose data that users have asked to be logged is behaviour that most will find surprising. IMO, they will either want metric data logged, or they won't – and I'd assume most won't. If you'll permit me to stand on a soapbox and preach for a while...

\<soapbox\>

> There's a fundamental difference between logging and metrics which the Pivotal folks have blurred with their choice to push metrics through the loggregator. If you google "logs and metrics" or "logs vs metrics" you'll find no end of debate around this, but most people write approximately this:
> 
> Each log message at severity > INFO should represent a single concrete Thing happening (e.g. an error occuring, a user request being served), and thus has the potential to be unique. This means they should be stored verbatim, without redaction (except of personally-identifiable information for long-term log storage), deduplication or aggregation. Performing any of these operations risks throwing away unique data, which you can't then see when looking at the logs after the fact.
> 
> Metrics, on the other hand, are timeseries data. That is: they represent repeated measurements of the same underlying value(s) over time. Aggregating and deduplicating this kind of data is not only OK, it's pretty much essential to extract small pieces of meaningful high-level value from the voluminous quantities of low-level metric data produced by any complex system.

\</soapbox\>

It's possible and sometimes useful to convert logs into metrics / timeseries for the purpose of visualisation by sampling and exporting counters / gauges. Conversely, the value of converting metrics into logs is not clear to me. A timeseries point is not individually useful – there's basically no use-case for finding exactly one emitted metric. I don't see any use for writing nozzle-derived counter metrics as logs for similar reasons, because the process of deriving the metrics throws away the raw data, and that's generally what should be logged.

The only advantage to logging all metrics that I can think of is to reconstruct a finer-grained view of a certain short timespan of history after the fact, because the aggregation done when collecting and storing the metric data as timeseries happened to throw away important information. For this to be feasible, the nozzle must treat logged metrics like any other log messages: potentially unique and thus not safe to dedupe or discard. But the cost-benefit of this is low, so I don't expect many people to choose to do this, and the nozzle should not default to doing so.

With that said, here's how I suggest we proceed:

*   Don't log "metric types" by default,
    *   i.e. keep firehose_events_to_stackdriver_logging set as is, and
    *   break the link from MetricRouter to LogAdapter
*   When users request "metric types" be logged too, send them through the logs sink verbatim.
*   Dig into the nozzle unreliability caused by logging metrics, that sounds like it needs fixing.

johnsonj:
Answer for this and #142 since they're tightly related. I'm sold and fully agree. Logs through the sampling pipeline is a hack/bandaid. Let's fix whatever real problem lies there if we find one. Thank you @knyar and @fluffle for walking me through all of this.

johnsonj:
Reviewed 11 of 12 files at r1.
Review status: 11 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/stackdriver-nozzle/heartbeat/metric_handler.go, line 82 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/156#-KzWtf47mPx2sZOio6Y3:-KzWtf47mPx2sZOio6Y4:b7ax4xt) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/034318fd01ee1c2d44ec4e749db00ca0af3aa98e/src/stackdriver-nozzle/heartbeat/metric_handler.go#L82)):*
> ```Go
> 		})
> 	}
> 	h.ma.PostMetrics(metrics)
> ```

woohoo this code is gone now :) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/156)*

johnsonj:
:lgtm: I think this is progress towards #142 but will require rebasing in the telemetry changes. 

---

Reviewed 1 of 12 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/156#-:-KzWuvZWR1RA2bDoVgd-:b-i60qsi)*

knyar:
Rebased.

johnsonj:
Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/156)*

johnsonj:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/156#-:-KzzTsaUnlTaUAoFsdRp:bnfp4nl)*

PR #160, opened 2017-11-27
Telemetry data under multiple prefixes.

fluffle:
Add another layer of abstraction, to allow telemetry counters to be
exported under prefixes other than stackdriver-nozzle. This paves
the way towards exporting nozzle-derived metrics based on firehose
data.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/160)


fluffle:
Hi Jeff, 

This is the first step along the journey towards having the Nozzle export per-application HTTP metrics derived from HttpStartStop events. I went for a minimal change here, so I've kept the previous exported functions built on top of the new structs. If you'd prefer other approaches just let me know.

Cheers,
--alex

johnsonj:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L--S7s3nN42gJqqjF0H:b-b3gmcl) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*
> ```Go
> )
> 
> const prefix = "stackdriver-nozzle/"
> ```

Essentially we're creating all sorts of data structures and factory methods to handle this string right here. We will maybe have 2 prefixes? `stackdriver-nozzle/` and `<user provided but firehose by default>/`

I'd rather see either:
- expose 'const Prefix', prepend it to all metric names that rely on this default
- an [options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) style API for specifying prefix, eg:

  ```go
  internalCounter = telemetry.NewCounter("internal")
  firehoseMetricCounter = telemetry.NewCounter("internal", telemetry.Prefix("firehose"))
  ```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

fluffle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-0pqmVqSFBXzeZ3rJe:b1lzpvc) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

Essentially we're creating all sorts of data structures and factory methods to handle this string right here. We will maybe have 2 prefixes? `stackdriver-nozzle/` and `<user provided but firehose by default>/`

I'd rather see either:
- expose 'const Prefix', prepend it to all metric names that rely on this default
- an [options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) style API for specifying prefix, eg:

  ```go
  internalCounter = telemetry.NewCounter("internal")
  firehoseMetricCounter = telemetry.NewCounter("internal", telemetry.Prefix("firehose"))
  ```
</blockquote></details>

The intent was to have >1, yes. I probably should have provided more context than the commit description.

I think it is important to distinguish between metrics that have come from the firehose directly and those that have been derived from that data, for the same reason that we distinguish between firehose data and the nozzle's internal metrics. Derived data requires Counters (and probably Gauges and other things) to track their values over time, so re-using the telemetry code seemed like the right thing to do.

I haven't completely firmed up the approach I want to take for the HTTP stats, but in terms of prefix either there would either be one prefix like "nozzle-derived" or "firehose-derived", or many -- this one would be something like "app-http-stats". Naming is hard :-(

I may be over-engineering for something that is demonstrating a proof-of-concept that may be better achieved in the gorouter. I think your criticism of there being way too much structure is entirely valid. But! The options syntax requires explicitly specifying the prefix for each metric, so it doesn't make much sense to me: the point of separating prefix from metric name is so that we don't have to do that. 

I will rework this today and get back to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

fluffle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-13v4k1ILmszaU3aVY:bgxjs2d) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

The intent was to have >1, yes. I probably should have provided more context than the commit description.

I think it is important to distinguish between metrics that have come from the firehose directly and those that have been derived from that data, for the same reason that we distinguish between firehose data and the nozzle's internal metrics. Derived data requires Counters (and probably Gauges and other things) to track their values over time, so re-using the telemetry code seemed like the right thing to do.

I haven't completely firmed up the approach I want to take for the HTTP stats, but in terms of prefix either there would either be one prefix like "nozzle-derived" or "firehose-derived", or many -- this one would be something like "app-http-stats". Naming is hard :-(

I may be over-engineering for something that is demonstrating a proof-of-concept that may be better achieved in the gorouter. I think your criticism of there being way too much structure is entirely valid. But! The options syntax requires explicitly specifying the prefix for each metric, so it doesn't make much sense to me: the point of separating prefix from metric name is so that we don't have to do that. 

I will rework this today and get back to you.


</blockquote></details>

How do you feel about this approach?

Usage would look like: 

appPrefix := telemetry.MetricPrefix("app-httpstartstop")
appRequests := appPrefix.NewCounter("requests")
appResponseCodes := appPrefix.NewCounterMap("response_codes")

metricSet could be decomposed into two package-level vars as before, I guess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

johnsonj:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-7jSjJokf3A6Nd1RRv:b-tp1aax) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

How do you feel about this approach?

Usage would look like: 

appPrefix := telemetry.MetricPrefix("app-httpstartstop")
appRequests := appPrefix.NewCounter("requests")
appResponseCodes := appPrefix.NewCounterMap("response_codes")

metricSet could be decomposed into two package-level vars as before, I guess.
</blockquote></details>

It's not different though- is it? We need to knowledge of the appPrefix for each version. IMO the options approach is more go-like and keeps the struct's POD.

```go
appPrefix := telemetry.MetrixPrefix("app-httpstartstop")
appRequests := appPrefix.NewCounter("requests")
// vs
appPrefix := telemetry.MetrixPrefix("app-httpstartstop")
appRequests :=telemetry.NewCounter("requests", telemetry.Prefix(appPrefix))
```

What I care about the most here is not duplicating any of the existing methods (`Do,Get,New*`) and creating a special case code path for nozzle metrics vs non-nozzle. If you want to accomplish that with the metricPrefix as a builder then I recommend dropping the package level version and expose `var StackdriverNozzle MetricType`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

fluffle:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-BCJDDaDeQM2Y7CcQJ:bblf5an) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

It's not different though- is it? We need to knowledge of the appPrefix for each version. IMO the options approach is more go-like and keeps the struct's POD.

```go
appPrefix := telemetry.MetrixPrefix("app-httpstartstop")
appRequests := appPrefix.NewCounter("requests")
// vs
appPrefix := telemetry.MetrixPrefix("app-httpstartstop")
appRequests :=telemetry.NewCounter("requests", telemetry.Prefix(appPrefix))
```

What I care about the most here is not duplicating any of the existing methods (`Do,Get,New*`) and creating a special case code path for nozzle metrics vs non-nozzle. If you want to accomplish that with the metricPrefix as a builder then I recommend dropping the package level version and expose `var StackdriverNozzle MetricType`
</blockquote></details>

That makes sense. I kept the package-level functions because it involved less refactoring, so the diff had a higher SNR. I've added a second commit on top of the first in this PR so that there are no longer package-level functions in telemetry. The refactoring made it a little clearer that counters should be exported for testing -- they are de-facto public anyway because they can be retrieved via the telemetry package.

It seems I misunderstood your suggestion of the Options approach, I was mostly averse to repeating constant strings for each metric. I still think it's overly complex for what's needed here. If we had more than one option, or if the prefix was in fact optional, then sure. But the prefix isn't really optional here, and the infrastructure for options processing feels a bit like my first pass at this change: lots of functions to handle a string. This way it really is just a string. But if you still feel strongly about it, I will take that approach.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

johnsonj:
Review status: 0 of 13 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go, line 32 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DFA8PEMyZ19_kzjnw:-L-DFA8QzT34hYxVv3zj:bmd7o2s) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go#L32)):*
> ```Go
> 
> var (
> 	// Exported so that tests outside the metrics_pipeline package can read values.
> ```

I don't see any tests outside of metrics_pipeline that need this. Maybe I'm missing it- I did just have eye surgery (hehe).

Instead of exporting all of the counters (which I agree is the same as it used to be), I'd prefer a change to the tests to be in the same package (s/metrics_pipeline_test/metrics_pipeline). If you don't want to do all that work in this change that's fine. 

---

*[src/stackdriver-nozzle/nozzle/nozzle_test.go, line 51 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DE9Hx3FDmN-NeW-bI:-L-DE9Hx3FDmN-NeW-bJ:bs2grt7) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/nozzle/nozzle_test.go#L51)):*
> ```Go
> 
> 	It("updates the counter", func() {
> 		nozzle.FirehoseEventsTotal.Set(0)
> ```

suggest moving to the BeforeEach block, If that seems too broad then put in its own Context("telemetry tests", func() {..) and do it there.

---

*[src/stackdriver-nozzle/stackdriver/metric_adapter_test.go, line 240 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DFgXGLMEM-xIgeqzv:-L-DFgXGLMEM-xIgeqzw:b-ghuyi6) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/stackdriver/metric_adapter_test.go#L240)):*
> ```Go
> 
> 	It("increments metrics counters", func() {
> 		stackdriver.FirehoseEventsCount.Set(0)
> ```

This should be in a BeforeEach somewhere (same comment to other affected tests)

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-DGC93moJ8HUZRbA28:bpjft2r) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

That makes sense. I kept the package-level functions because it involved less refactoring, so the diff had a higher SNR. I've added a second commit on top of the first in this PR so that there are no longer package-level functions in telemetry. The refactoring made it a little clearer that counters should be exported for testing -- they are de-facto public anyway because they can be retrieved via the telemetry package.

It seems I misunderstood your suggestion of the Options approach, I was mostly averse to repeating constant strings for each metric. I still think it's overly complex for what's needed here. If we had more than one option, or if the prefix was in fact optional, then sure. But the prefix isn't really optional here, and the infrastructure for options processing feels a bit like my first pass at this change: lots of functions to handle a string. This way it really is just a string. But if you still feel strongly about it, I will take that approach.
</blockquote></details>

OK with me. 

Prefix would be optional if we default it to `stackdriver-nozzle` but perhaps that's not good. I think a package level variable to contain the name as you've done here is a nice solution. 

I don't think the naming is right, "a MetricPrefix builds a Counter" which makes me hesitant about the API. That's why I lean towards the prefix being part of the creation call "the telemetry package creates a Counter with a MetricPrefix and Name". 

I think it's structurally sound and will accomplish what we both want so let's move forward.  The removal of the telemetrytest package to me is a good sign that this is progress.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

fluffle:
Review status: 0 of 23 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go, line 32 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DFA8PEMyZ19_kzjnw:-L-GTIQSWnQZw7ee_OPk:blq6724) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go#L32)):*

I don't see any tests outside of metrics_pipeline that need this. Maybe I'm missing it- I did just have eye surgery (hehe).

Instead of exporting all of the counters (which I agree is the same as it used to be), I'd prefer a change to the tests to be in the same package (s/metrics_pipeline_test/metrics_pipeline). If you don't want to do all that work in this change that's fine. 
</blockquote></details>

Sure, I can do that. I thought there were gingko-related reasons as to why metrics_pipeline and metrics_pipeline_test were separate packages, that's why I didn't take this approach at first. 

---

*[src/stackdriver-nozzle/nozzle/nozzle_test.go, line 51 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DE9Hx3FDmN-NeW-bI:-L-G_q4AlExTr972ru36:bud1cqi) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/nozzle/nozzle_test.go#L51)):*

suggest moving to the BeforeEach block, If that seems too broad then put in its own Context("telemetry tests", func() {..) and do it there.
</blockquote></details>

Done, I should have done that to begin with sorry!

---

*[src/stackdriver-nozzle/stackdriver/metric_adapter_test.go, line 240 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-DFgXGLMEM-xIgeqzv:-L-GcVVLYpJfkUU5uUVX:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/47c7b52a3c9e01fc184c33afa190c34945e02237/src/stackdriver-nozzle/stackdriver/metric_adapter_test.go#L240)):*

This should be in a BeforeEach somewhere (same comment to other affected tests)
</blockquote></details>

Done.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-GTi3B9OPiWKZOw4I-:b-e7ht1l) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

OK with me. 

Prefix would be optional if we default it to `stackdriver-nozzle` but perhaps that's not good. I think a package level variable to contain the name as you've done here is a nice solution. 

I don't think the naming is right, "a MetricPrefix builds a Counter" which makes me hesitant about the API. That's why I lean towards the prefix being part of the creation call "the telemetry package creates a Counter with a MetricPrefix and Name". 

I think it's structurally sound and will accomplish what we both want so let's move forward.  The removal of the telemetrytest package to me is a good sign that this is progress.
</blockquote></details>

Hmmm. I see your point about the naming. And I am aware I have a weakness for putting methods on things :-)

It's a minor change to reinstate the package-level New functions and require a MetricPrefix as their first argument -- after all that's basically what happens under the hood anyway, it's just a question of how you want the syntax to look. We go from:

```Go
counter := telemetry.Nozzle.NewCounter("name")
```

to

```Go
counter := telemetry.NewCounter(telemetry.Nozzle, "name")
```

The latter is very similar to the Option approach you proposed but without the ... optionality? :-D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160)*

johnsonj:
:lgtm: awesome work @fluffle, between this CR and your help with #152 this has been a huge leap for the nozzle in metric usability and code health. 

---

Reviewed 2 of 13 files at r3, 21 of 21 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/stackdriver-nozzle/cmd/firehose_rate.go, line 58 at r4](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-J-c3Cf2cFuq2CqKFU:-L-J-c3Cf2cFuq2CqKFV:b-yjsczp) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/06cfa8613406ec1c5bb2ff18136472b4f602070c/src/stackdriver-nozzle/cmd/firehose_rate.go#L58)):*
> ```Go
> 	defer counter.Stop()
> 	for _ = range messages {
> 		counter.Increment("count")
> ```

forgot about this code! thanks for find/fix :) 

---

*[src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go, line 36 at r4](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-J04kFPZL81Rw3ac-A:-L-J04kFPZL81Rw3ac-B:bg8mocj) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/68cd08d6ad563fded66e4a6d3a523dfff07abcfd/src/stackdriver-nozzle/metrics_pipeline/auto_culled_metrics_buffer.go#L36)):*
> ```Go
> 
> func init() {
> 	eventsSampledCount = telemetry.NewCounter(telemetry.Nozzle, "metrics.firehose_events.sampled.count")
> ```

this looks great! 

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 24 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L--S7s3nN42gJqqjF0G:-L-J0lNFq24ekYxj5aXf:b-z2p2cq) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/13615be557c98c9637594803e75227becff87bd2/src/stackdriver-nozzle/telemetry/counters.go#L24)):*

Hmmm. I see your point about the naming. And I am aware I have a weakness for putting methods on things :-)

It's a minor change to reinstate the package-level New functions and require a MetricPrefix as their first argument -- after all that's basically what happens under the hood anyway, it's just a question of how you want the syntax to look. We go from:

```Go
counter := telemetry.Nozzle.NewCounter("name")
```

to

```Go
counter := telemetry.NewCounter(telemetry.Nozzle, "name")
```

The latter is very similar to the Option approach you proposed but without the ... optionality? :-D
</blockquote></details>

Awesome awesome awesome. 

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 32 at r4](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-L-J0yoJtVWMNieSENEw:-L-J0yoJtVWMNieSENEx:b-dh1u9a) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/68cd08d6ad563fded66e4a6d3a523dfff07abcfd/src/stackdriver-nozzle/telemetry/counters.go#L32)):*
> ```Go
> }
> 
> // IntValue returns the counter's value as an int rather than an int64. Tests are
> ```

agreed this is needed. I personally think it's a shame ginkgo/gomega doesn't match `Expect(int64(5)).To(Be(5))` given 5 is a constant. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/160#-:-L-J1Qu4c0ZX7_lVjFHx:b-8ihfqf)*

PR #162, opened 2017-12-06
Report CounterEvents as cumulative metrics to Stackdriver

knyar:
This adds a "counter tracker" which is used to maintain start time and a separate cumulative value for each counter metric, which need to be reported to Stackdriver.

As the result, each counter is properly exported as a cumulative Stackdriver metric instead of two gauge metrics (`.total` and `.delta`).

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/162)

### DIFF
--- a/jobs/stackdriver-nozzle/spec
+++ b/jobs/stackdriver-nozzle/spec
@@ -67,3 +67,7 @@ properties:
   nozzle.foundation_name:
     description: Name added as the 'foundation' label to all time series being sent to Stackdriver, useful for differentiating between multiple PCF instances in a project.
     default: cf
+
+  nozzle.enable_cumulative_counters:
+    description: Enable reporting counter events as cumulative Stackdriver metrics. This requires all CounterEvent messages for a given metric to be routed to the same nozzle process (which is the case if you run a single copy of the nozzle).
+    default: false
--- a/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
+++ b/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
@@ -38,6 +38,7 @@ case $1 in
     export FOUNDATION_NAME=<%= p('nozzle.foundation_name', 'cf') %>
     export LOGGING_BATCH_COUNT=<%= p('nozzle.logging_batch_count', '1000') %>
     export LOGGING_BATCH_DURATION=<%= p('nozzle.logging_batch_duration', '30') %>
+    export ENABLE_CUMULATIVE_COUNTERS=<%= p('nozzle.enable_cumulative_counters', 'false') %>
 
     <% if_p('gcp.project_id') do |prop| %>
     export GCP_PROJECT_ID=<%= prop %>
--- a/src/stackdriver-nozzle/app/builder.go
+++ b/src/stackdriver-nozzle/app/builder.go
@@ -129,7 +129,13 @@ func (a *App) newMetricSink(ctx context.Context, metricAdapter stackdriver.Metri
 	metricBuffer := metrics_pipeline.NewAutoCulledMetricsBuffer(ctx, a.logger, time.Duration(a.c.MetricsBufferDuration)*time.Second, metricAdapter)
 	a.bufferEmpty = metricBuffer.IsEmpty
 
-	return nozzle.NewMetricSink(a.logger, a.c.MetricPathPrefix, a.labelMaker, metricBuffer, nozzle.NewUnitParser(), a.c.RuntimeMetricRegex)
+	var counterTracker *nozzle.CounterTracker
+	if a.c.EnableCumulativeCounters {
+		ttl := time.Duration(a.c.CounterTrackerTTL) * time.Second
+		counterTracker = nozzle.NewCounterTracker(ctx, ttl, a.logger)
+	}
+
+	return nozzle.NewMetricSink(a.logger, a.c.MetricPathPrefix, a.labelMaker, metricBuffer, counterTracker, nozzle.NewUnitParser(), a.c.RuntimeMetricRegex)
 }
 
 func (a *App) newTelemetryReporter() telemetry.Reporter {
--- a/src/stackdriver-nozzle/config/config.go
+++ b/src/stackdriver-nozzle/config/config.go
@@ -74,6 +74,13 @@ type Config struct {
 	DebugNozzle           bool   `envconfig:"debug_nozzle"`
 	// By default 'origin' label is prepended to metric name, however for runtime metrics (defined here) we add it as a metric label instead.
 	RuntimeMetricRegex string `envconfig:"runtime_metric_regex" default:"^(numCPUS|numGoRoutines|memoryStats\\..*)$"`
+	// If enabled, CounterEvents will be reported as cumulative Stackdriver metrics instead of two gauges (<metric>.delta
+	// and <metric>.total). Reporting cumulative metrics involves nozzle keeping track of internal counter state, and
+	// requires deterministic routing of CounterEvents to nozzles (i.e. CounterEvent messages for a particular metric MUST
+	// always be routed to the same nozzle process); the easiest way to achieve that is to run a single copy of the nozzle.
+	EnableCumulativeCounters bool `envconfig:"enable_cumulative_counters"`
+	// Expire internal counter state if a given counter has not been seen for this many seconds.
+	CounterTrackerTTL int `envconfig:"counter_tracker_ttl" default:"130"`
 }
 
 func (c *Config) validate() error {
--- a/src/stackdriver-nozzle/messages/metric.go
+++ b/src/stackdriver-nozzle/messages/metric.go
@@ -2,10 +2,16 @@ package messages
 
 import (
 	"bytes"
+	"fmt"
+	"path"
 	"sort"
 	"time"
 
 	"github.com/cloudfoundry/sonde-go/events"
+	"github.com/golang/protobuf/ptypes/timestamp"
+	labelpb "google.golang.org/genproto/googleapis/api/label"
+	metricpb "google.golang.org/genproto/googleapis/api/metric"
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 )
 
 // Metric represents one of the metrics contained in an events.Envelope.
@@ -13,25 +19,96 @@ type Metric struct {
 	Name      string
 	Labels    map[string]string `json:"-"`
 	Value     float64
+	IntValue  int64
 	EventTime time.Time
+	StartTime time.Time                 `json:"-"`
 	Unit      string                    // TODO Should this be "1" if it's empty?
 	Type      events.Envelope_EventType `json:"-"`
+}
+
+func (m *Metric) IsCumulative() bool {
+	return m.Type == events.Envelope_CounterEvent
+}
+
+func (m *Metric) metricType() string {
+	return path.Join("custom.googleapis.com", m.Name)
+}
+
+// NeedsMetricDescriptor determines whether a custom metric descriptor needs to be created for this metric in Stackdriver.
+// We do that if we need to set a custom unit, or mark metric as a cumulative.
+func (m *Metric) NeedsMetricDescriptor() bool {
+	return m.Unit != "" || m.IsCumulative()
+}
+
+// MetricDescriptor returns a Stackdriver MetricDescriptor proto for this metric.
+func (m *Metric) MetricDescriptor(projectName string) *metricpb.MetricDescriptor {
+	metricType := m.metricType()
+
+	var labelDescriptors []*labelpb.LabelDescriptor
+	for key := range m.Labels {
+		labelDescriptors = append(labelDescriptors, &labelpb.LabelDescriptor{
+			Key:       key,
+			ValueType: labelpb.LabelDescriptor_STRING,
+		})
+	}
+
+	metricKind := metricpb.MetricDescriptor_GAUGE
+	valueType := metricpb.MetricDescriptor_DOUBLE
+	if m.IsCumulative() {
+		metricKind = metricpb.MetricDescriptor_CUMULATIVE
+		valueType = metricpb.MetricDescriptor_INT64
+	}
+
+	return &metricpb.MetricDescriptor{
+		Name:        path.Join(projectName, "metricDescriptors", metricType),
+		Type:        metricType,
+		Labels:      labelDescriptors,
+		MetricKind:  metricKind,
+		ValueType:   valueType,
+		Unit:        m.Unit,
+		Description: "stackdriver-nozzle created custom metric.",
+		DisplayName: m.Name,
+	}
+}
+
+// TimeSeries returns a Stackdriver TimeSeries proto for this metric value.
+func (m *Metric) TimeSeries() *monitoringpb.TimeSeries {
+	var value *monitoringpb.TypedValue
+	if m.IsCumulative() {
+		value = &monitoringpb.TypedValue{Value: &monitoringpb.TypedValue_Int64Value{Int64Value: m.IntValue}}
+	} else {
+		value = &monitoringpb.TypedValue{Value: &monitoringpb.TypedValue_DoubleValue{DoubleValue: m.Value}}
+	}
+
+	point := &monitoringpb.Point{
+		Interval: &monitoringpb.TimeInterval{
+			EndTime:   &timestamp.Timestamp{Seconds: m.EventTime.Unix(), Nanos: int32(m.EventTime.Nanosecond())},
+			StartTime: &timestamp.Timestamp{Seconds: m.StartTime.Unix(), Nanos: int32(m.StartTime.Nanosecond())},
+		},
+		Value: value,
+	}
+	return &monitoringpb.TimeSeries{
+		Metric: &metricpb.Metric{
+			Type:   m.metricType(),
+			Labels: m.Labels,
+		},
+		Points: []*monitoringpb.Point{point},
+	}
 }
 
 func (m *Metric) Hash() string {
 	var b bytes.Buffer
 
-	// Extract keys to a slice and sort it
-	numKeys := len(m.Labels) + 1
-	keys := make([]string, numKeys, numKeys)
-	keys = append(keys, m.Name)
+	b.Write([]byte(m.Name))
+
+	// Extract label keys to a slice and sort it
+	keys := make([]string, 0, len(m.Labels))
 	for k := range m.Labels {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 	for _, k := range keys {
-		b.Write([]byte(k))
-		b.Write([]byte(m.Labels[k]))
+		b.Write([]byte(fmt.Sprintf(",%s='%s'", k, m.Labels[k])))
 	}
 	return b.String()
 }
--- a/src/stackdriver-nozzle/nozzle/counter_tracker.go
+++ b/src/stackdriver-nozzle/nozzle/counter_tracker.go
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nozzle
+
+import (
+	"context"
+	"expvar"
+	"math"
+	"sync"
+	"time"
+
+	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/telemetry"
+	"github.com/cloudfoundry/lager"
+)
+
+const maxExpirePeriod = 10 * time.Second
+
+var countersExpiredCount *telemetry.Counter
+
+func init() {
+	countersExpiredCount = telemetry.NewCounter(telemetry.Nozzle, "metrics.counters.expired")
+}
+
+type counterData struct {
+	startTime     time.Time
+	totalValue    *expvar.Int
+	lastValue     uint64
+	lastSeenTime  time.Time
+	lastEventTime time.Time
+}
+
+// CounterTracker is used to provide a "start time" for each loggregator counter metric exported by the nozzle.
+//
+// Stackdriver requires each point for a cumulative metric to include "start time" in addition to the actual event time
+// (aka "end time"): https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#point
+// Typically start time would correspond to the time when the actual process exporting the metric started. This ensures
+// that when a process is restarted (and counter gets reset to 0), start time increases.
+//
+// Since binaries that export counter events to loggregator only provide event time, the nozzle needs to determine start
+// time for each metric itself. To do that, CounterTracker keeps its own counter for each metric, which corresponds to the
+// total number of events since the metric was first seen by the nozzle (which is exported as the start time).
+//
+// As an example, a series of incoming CounterEvents with total values of [100, 110, 115, 150] will be exported by the
+// nozzle as [10, 15, 50] (first point seen by the nozzle is discarded, because each point reported to Stackdriver needs
+// to cover non-zero time interval between start time and end time).
+//
+// If CounterTracker detects the total value for a given counter decrease, it will interpret this as a counter reset. This
+// will not result in the Stackdriver cumulative metric being reset as well; for example, incoming CounterEvents with total
+// values of [100, 110, 115, 10, 17] will be exported by the nozzle as [10, 15, 25, 32].
+//
+// CounterTracker will regularly remove internal state for metrics that have not been seen for a while. This is done to
+// conserve memory, and also to ensure that old values do not re-surface if a given counter stops being exported for some
+// period of time.
+type CounterTracker struct {
+	counters map[string]*counterData
+	mu       *sync.Mutex // protects `counters`
+	ttl      time.Duration
+	logger   lager.Logger
+	ticker   *time.Ticker
+	ctx      context.Context
+}
+
+// NewCounterTracker creates and returns a counter tracker.
+func NewCounterTracker(ctx context.Context, ttl time.Duration, logger lager.Logger) *CounterTracker {
+	expirePeriod := time.Duration(ttl.Nanoseconds() / 2)
+	if expirePeriod > maxExpirePeriod {
+		expirePeriod = maxExpirePeriod
+	}
+	c := &CounterTracker{
+		counters: map[string]*counterData{},
+		mu:       &sync.Mutex{},
+		ttl:      ttl,
+		logger:   logger,
+		ticker:   time.NewTicker(expirePeriod),
+		ctx:      ctx,
+	}
+	go func() {
+		for {
+			select {
+			case <-c.ticker.C:
+				c.expire()
+			case <-c.ctx.Done():
+				c.ticker.Stop()
+				return
+			}
+		}
+	}()
+	return c
+}
+
+// Update accepts a counter name, event time and a value, and returns the total value for the counter along with its
+// start time. Counter name provided needs to uniquely identify the time series (so it needs to include metric name as
+// well as all metric label values).
+// At least two values need to be observed for a given counter to determine the total value, so for the first observed
+// value, 0 will be returned as the total, and end time will be equal to event time. Such points should not be reported
+// to Stackdriver, since it expects points covering non-zero time interval.
+func (t *CounterTracker) Update(name string, value uint64, eventTime time.Time) (int64, time.Time) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	c, present := t.counters[name]
+	if !present {
+		c = t.newCounterData(name, eventTime)
+		t.counters[name] = c
+	} else {
+		var delta uint64
+		if c.lastValue > value {
+			// Counter has been reset.
+			delta = value
+		} else {
+			delta = value - c.lastValue
+		}
+		if uint64(c.totalValue.Value())+delta > math.MaxInt64 {
+			// Accumulated value overflows int64, we need to reset the counter.
+			c.totalValue.Set(int64(delta))
+			c.startTime = c.lastEventTime
+		} else {
+			c.totalValue.Add(int64(delta))
+		}
+	}
+	c.lastValue = value
+	c.lastSeenTime = time.Now()
+	c.lastEventTime = eventTime
+	return c.totalValue.Value(), c.startTime
+}
+
+func (t *CounterTracker) newCounterData(name string, eventTime time.Time) *counterData {
+	var v *expvar.Int
+	existing := expvar.Get(name)
+	if existing != nil {
+		// There was a previous counter with this name; use it instead, but reset value to 0.
+		v = existing.(*expvar.Int)
+		v.Set(0)
+	} else {
+		v = expvar.NewInt(name)
+	}
+	// Initialize counter state for a new counter.
+	return &counterData{
+		totalValue: v,
+		startTime:  eventTime,
+	}
+}
+
+func (t *CounterTracker) expire() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	for name, counter := range t.counters {
+		if time.Now().Sub(counter.lastSeenTime) > t.ttl {
+			t.logger.Info("CounterTracker", lager.Data{
+				"info":    "removing expired counter",
+				"name":    name,
+				"counter": counter,
+				"value":   t.counters[name].totalValue.Value(),
+			})
+			// Reset values to -1 to make expired counters visible in /debug/vars.
+			t.counters[name].totalValue.Set(-1)
+			delete(t.counters, name)
+			countersExpiredCount.Increment()
+		}
+	}
+}
--- a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go
+++ b/src/stackdriver-nozzle/nozzle/counter_tracker_test.go
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nozzle
+
+import (
+	"context"
+	"math"
+	"time"
+
+	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/mocks"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func testCounterTracker(subject *CounterTracker, name string, baseTime time.Time, incoming []uint64, expected []int64) {
+	for idx, value := range incoming {
+		ts := baseTime.Add(time.Duration(idx) * time.Millisecond)
+		total, st := subject.Update(name, value, ts)
+		if idx == 0 {
+			// First seen value initializes the counter.
+			Expect(total).To(BeNumerically("==", 0))
+		} else {
+			Expect(total).To(BeNumerically("==", expected[idx-1]))
+			Expect(st).To(BeTemporally("~", baseTime))
+		}
+	}
+}
+
+var _ = Describe("CounterTracker", func() {
+	var (
+		subject    *CounterTracker
+		counterTTL time.Duration
+		logger     *mocks.MockLogger
+	)
+
+	BeforeEach(func() {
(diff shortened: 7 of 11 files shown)


johnsonj:
:lgtm: structure is good, just nit stuff.

With this change do we need to say only run a single instance of the nozzle (until consistent nozzling lands)? 

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/stackdriver-nozzle/messages/metric.go, line 31 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hc3LX6N9peM0nAWl6:-L-hc3LX6N9peM0nAWl7:bb65fny) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/messages/metric.go#L31)):*
> ```Go
> 	// Extract keys to a slice and sort it
> 	numKeys := len(m.Labels) + 1
> 	keys := make([]string, 0, numKeys)
> ```

good catch since we're appending below. 

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 101 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hdg4kGqn74UIkIq_4:-L-hdg4kGqn74UIkIq_5:bgzhgun) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L101)):*
> ```Go
> // At least two values need to be observed for a given counter to determine the total value, so for the first observed
> // value, `nil` will be returned instead of the start time. In that case, the returned value should be discarded.
> func (t *counterTracker) GetTotal(name string, value float64, eventTime *time.Time) (float64, *time.Time) {
> ```

suggest naming: UpdateCounter, Track, Update, ... Get isn't generally stateful. 

Curious why you went with passing/returning eventTime as a pointer?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 114 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-he3AaDnOUKH3AZS11:-L-he3AaDnOUKH3AZS12:bc66gz6) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L114)):*
> ```Go
> 			lastSeenTime: time.Now(),
> 		}
> 		return 0, nil
> ```

suggest adding another return parameter, `ok bool`. I don't think the caller should make the decision to publish the metric based on the metric/time value. 

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 117 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hfYegkUZLozg4V4_j:-L-hfYegkUZLozg4V4_k:b-ok3vtf) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L117)):*
> ```Go
> 	}
> 	if c.lastValue > value {
> 		// Counter has been reset.
> ```

is there a reason not to reset our counter as well in this scenario? 

---

*[src/stackdriver-nozzle/nozzle/metric_sink_test.go, line 241 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hhkp6EGFomin59uFX:-L-hhkp7SysDbSutyp4y:b71hxym) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/metric_sink_test.go#L241)):*
> ```Go
> 		expectedTotals := []float64{10, 20, 25, 45}
> 		for idx, total := range expectedTotals {
> 			Expect(metrics[idx]).To(MatchFields(IgnoreExtras, Fields{
> ```

`IgnoreExtras`- very awesome. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L-his8TJ4sfg5CCCtTa:b-n19tlj)*

fluffle:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 31 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-katU30aKeFvHQDJul:-L-katU30aKeFvHQDJum:bpogukh) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L31)):*
> ```Go
> 
> func init() {
> 	countersExpiredCount = telemetry.NewCounter(telemetry.Nozzle, "metrics.counters.expired.count")
> ```

Naming nit: "total" instead of "count" for consistency? or just drop the "count" (it's cleaner!)

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 66 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kj8Z5g2pRqyzSYnVf:-L-kj8Z5g2pRqyzSYnVg:b-9ghocf) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L66)):*
> ```Go
> 	counters map[string]*counterData
> 	mu       *sync.Mutex // protects `counters`
> 	ttl      *time.Duration
> ```

Similarly here time.Duration is really just an int64 so this should probably not be a pointer. I can't find any "pointer is nil" overloading for this particular field either.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 79 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kdYDlgkMRrmNNl2AD:-L-kdYDlgkMRrmNNl2AE:bieaqq0) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L79)):*
> ```Go
> 		ttl:      &ttl,
> 		logger:   logger,
> 		ticker:   time.NewTicker(time.Duration(ttl.Nanoseconds() / 2)),
> ```

Is this frequent enough? It means that the default TTL of 70s really means "between 70 and 105 seconds", which is a relatively large interval. 

I understand why basing it on the TTL makes sense -- to scale down to small TTLs effectively -- but maybe there needs to be an upper limit on the tick time to prevent large TTLs having a lot of variance?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 101 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hdg4kGqn74UIkIq_4:-L-kd-mQQwSVXUnjZLBK:b8ljxqf) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L101)):*

suggest naming: UpdateCounter, Track, Update, ... Get isn't generally stateful. 

Curious why you went with passing/returning eventTime as a pointer?
</blockquote></details>

+1 to time not being a pointer. https://golang.org/pkg/time/#Time says:

"Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time."

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 114 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-he3AaDnOUKH3AZS11:-L-kfxjuDUnAPTu1UopR:b-oiq46y) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L114)):*

suggest adding another return parameter, `ok bool`. I don't think the caller should make the decision to publish the metric based on the metric/time value. 
</blockquote></details>

I agree that a bool that indicates whether the metric should be published is a better approach than having the nil-ness of a time.Time pointer convey the same information.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 42 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-keetLbKsk4hW8NwUD:-L-keetM4TWw6kEw49ZJ:b-cxdv8f) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L42)):*
> ```Go
> 
> 	It("increments counters and handles counter resets", func() {
> 		ctx, cancel := context.WithCancel(context.TODO())
> ```

Isn't context.Background() OK to use in tests?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kfKTWRHX1XHZFnYcf:-L-kfKTWRHX1XHZFnYcg:begy8ht) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L69)):*
> ```Go
> 		subject = NewCounterTracker(ctx, counterTTL, logger)
> 
> 		runTest := func(baseTime time.Time) {
> ```

I think with a little effort you could significantly reduce the amount of copy/paste code here...

---

*[src/stackdriver-nozzle/nozzle/metric_sink_test.go, line 208 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kg_OBnJqyLRKQhwiS:-L-kg_OBnJqyLRKQhwiT:bu7z5y2) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/metric_sink_test.go#L208)):*
> ```Go
> 
> 		for idx, values := range eventValues {
> 			ts := eventTime.UnixNano() + int64(idx*1e9) // Events are 1 second apart.
> ```

Ugh, many integer types. I was going to say  that you should probably prefer to use time.Second here, but the amount of typecasting required for this to work is awkward. Even so,

eventTime.UnixNano() + int64(time.Second)*int64(idx)

or similar might be better.

---

*[src/stackdriver-nozzle/nozzle/metric_sink_test.go, line 209 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kiAatKcIyY8ztCzGC:-L-kiAatKcIyY8ztCzGD:b-5v3u07) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/metric_sink_test.go#L209)):*
> ```Go
> 		for idx, values := range eventValues {
> 			ts := eventTime.UnixNano() + int64(idx*1e9) // Events are 1 second apart.
> 			delta := uint64(values[0])
> ```

Is the cast required here? eventValues is [][]uint64 already.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162)*

knyar:
Oh yes, good point: this requires running a single copy of the nozzle. Should I document that somewhere, or maybe keep the old behavior making the counterTracker path enabled by a configuration option?

I have also discussed this with @fluffle this morning and made some significant changes:
- since counter values are uint64 in Firehose, I've changed cumulative metrics to be integers as well (int64). This required some wider changes, and creating Stackdriver protos seems to be more convenient in the Metric now. This seems like it also moves us slightly closer to #142.
- `expvar` is now used to store actual counters. I've changed `metric.Hash()` to be slightly more human readable to make metrics easy to read in `/debug/vars`.

---

Review status: 2 of 9 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 31 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-katU30aKeFvHQDJul:-L-l7g2SBhdx5QElFJis:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L31)):*

Naming nit: "total" instead of "count" for consistency? or just drop the "count" (it's cleaner!)
</blockquote></details>

Done.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 66 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kj8Z5g2pRqyzSYnVf:-L-l7zv-gy0fGiuPUk_p:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L66)):*

Similarly here time.Duration is really just an int64 so this should probably not be a pointer. I can't find any "pointer is nil" overloading for this particular field either.
</blockquote></details>

Done.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 79 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kdYDlgkMRrmNNl2AD:-L-l9QylSQ6bc6EpLVpz:b-5kyjf1) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L79)):*

Is this frequent enough? It means that the default TTL of 70s really means "between 70 and 105 seconds", which is a relatively large interval. 

I understand why basing it on the TTL makes sense -- to scale down to small TTLs effectively -- but maybe there needs to be an upper limit on the tick time to prevent large TTLs having a lot of variance?
</blockquote></details>

Good point, thanks. Capped at 10 seconds.
Also, default TTL increased to 130 seconds since some metrics are only emitted every minute and I'd like to ensure a single lost point does not result in a counter reset.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 101 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hdg4kGqn74UIkIq_4:-L-lJuavzW51G3mNiMAM:b-8etzn1) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L101)):*

+1 to time not being a pointer. https://golang.org/pkg/time/#Time says:

"Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time."
</blockquote></details>

Renamed to Update().

Switched time types to be passed as values. Instead of interpreting a nil pointer as indication of the first value seen for a counter, metric sink just uses the cumulative value of 0 (which seems more intuitive anyway).

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 114 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-he3AaDnOUKH3AZS11:-L-lKPNcfBfZkEL5oAbr:b-pycgw6) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L114)):*

I agree that a bool that indicates whether the metric should be published is a better approach than having the nil-ness of a time.Time pointer convey the same information.
</blockquote></details>

Currently that decision is made based on returned cumulative value being more than 0.

This seems logical, since the returned value is in fact correct and it *could* be used if Stackdriver accepted points where event time == start time. So the decision of whether to publish that value or not should be made by the caller, I think. Please let me know if you disagree.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 117 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hfYegkUZLozg4V4_j:-L-lLQHuVM2pBpYt7fK7:b-vof6p1) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L117)):*

is there a reason not to reset our counter as well in this scenario? 
</blockquote></details>

Well, I don't see a reason to reset it, given that we have all the state to continue adding to the existing total.
Also, I suspect detecting counter resets has a non-zero (albeit probably quite small) cost in Monarch.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 42 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-keetLbKsk4hW8NwUD:-L-lM0i9HahIKCKQhTGj:b-14wt0t) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L42)):*

Isn't context.Background() OK to use in tests?
</blockquote></details>

It does seem like it should be OK, however other tests are using context.TODO().
@johnsonj, any reason for that, or should I change the other tests as well (with a separate commit)?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kfKTWRHX1XHZFnYcf:-L-lqJjrJtX1a8vm3mbE:b8cbzsk) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L69)):*

I think with a little effort you could significantly reduce the amount of copy/paste code here...
</blockquote></details>

Done. Not sure if this is the right way to do such things in Ginkgo.

---

*[src/stackdriver-nozzle/nozzle/metric_sink_test.go, line 208 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kg_OBnJqyLRKQhwiS:-L-lr2KOLTVAR90ki201:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/metric_sink_test.go#L208)):*

Ugh, many integer types. I was going to say  that you should probably prefer to use time.Second here, but the amount of typecasting required for this to work is awkward. Even so,

eventTime.UnixNano() + int64(time.Second)*int64(idx)

or similar might be better.


</blockquote></details>

Done.

---

*[src/stackdriver-nozzle/nozzle/metric_sink_test.go, line 209 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kiAatKcIyY8ztCzGC:-L-lqn7UFv4trn5Dnpxd:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/metric_sink_test.go#L209)):*

Is the cast required here? eventValues is [][]uint64 already.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L-l7PRls24ptrtN5HtT:b-ptfsqo)*

johnsonj:
I think we should make it enabled by configuration. This is the ideal behavior but if we enforce single node deployments then we break scaling where it works fine today (for Stackdriver Logging).

---

Reviewed 7 of 7 files at r2.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/stackdriver-nozzle/messages/metric.go, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mp-bPCADvbyswXTz3:-L-mp-bPCADvbyswXTz4:b-o56v11) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/messages/metric.go#L38)):*
> ```Go
> 
> // NeedsMetricDescriptor determines whether a custom metric descriptor needs to be created for this metric in Stackdriver.
> // We do that if we need to set a custom unit, or mark metric as a cumulative.
> ```

very helpful comment!

---

*[src/stackdriver-nozzle/messages/metric.go, line 70 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mnCUDZ7DwQUzjR--n:-L-mnCUDZ7DwQUzjR--o:b13ewev) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/messages/metric.go#L70)):*
> ```Go
> 		Unit:        m.Unit,
> 		Description: "stackdriver-nozzle created custom metric.",
> 		DisplayName: m.Name, // TODO
> ```

is this TODO relevant or copy/pasted?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 79 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kdYDlgkMRrmNNl2AD:-L-mpFPG3x8b-OEYS9It:b-y1jwnr) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L79)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

Good point, thanks. Capped at 10 seconds.
Also, default TTL increased to 130 seconds since some metrics are only emitted every minute and I'd like to ensure a single lost point does not result in a counter reset.
</blockquote></details>

I think a fixed period would be just fine and more predicable (1? 5? 10 seconds?). Agree with the TTL increase being larger than two metric periods.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 101 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-hdg4kGqn74UIkIq_4:-L-moX4JvxjwudFB5i15:b-3ru0ze) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L101)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

Renamed to Update().

Switched time types to be passed as values. Instead of interpreting a nil pointer as indication of the first value seen for a counter, metric sink just uses the cumulative value of 0 (which seems more intuitive anyway).
</blockquote></details>

New header looks good. Resolving as the 'ok' discussion is below. 

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 114 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-he3AaDnOUKH3AZS11:-L-mnw6wiAYCrCcHZOwS:brojs4m) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L114)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

Currently that decision is made based on returned cumulative value being more than 0.

This seems logical, since the returned value is in fact correct and it *could* be used if Stackdriver accepted points where event time == start time. So the decision of whether to publish that value or not should be made by the caller, I think. Please let me know if you disagree.
</blockquote></details>

0 seems like a potential real value to me. If a counter was 'outOfMemoryErrors' then it very well may emit a cumulative 0 for a long period of time. I think we want to record that value in Stackdriver Monitoring and not hide it.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 165 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mqKynAikVOa-wSlEv:-L-mqKynAikVOa-wSlEw:b-mw6ehm) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/nozzle/counter_tracker.go#L165)):*
> ```Go
> 		if time.Now().Sub(counter.lastSeenTime) > t.ttl {
> 			t.counters[name].totalValue.Set(-1)
> 			t.logger.Info("counterTracker", lager.Data{
> ```

suggest logging before resetting the value. is it necessary to reset the value? it looks like newCounterData will take care of that if it reappears.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 42 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-keetLbKsk4hW8NwUD:-L-mqbnFmoRDk-NCDvGK:b-j1is2) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L42)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

It does seem like it should be OK, however other tests are using context.TODO().
@johnsonj, any reason for that, or should I change the other tests as well (with a separate commit)?

</blockquote></details>

agree Background is correct. Would appreciate that clean up in another commit. 

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kfKTWRHX1XHZFnYcf:-L-mqogTe9U9k9O6HZZH:b-nmfny) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L69)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

Done. Not sure if this is the right way to do such things in Ginkgo.
</blockquote></details>

It's totally reasonable and a good clean up. Ginkgo allows you to create custom matchers[1] but it may not be the right fit here.

[1] https://onsi.github.io/gomega/#adding-your-own-matchers

---

*[src/stackdriver-nozzle/stackdriver/metric_adapter.go, line 155 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mrbBA3GzAZbFTt9X7:-L-mrbBA3GzAZbFTt9X8:b-49y92c) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/stackdriver/metric_adapter.go#L155)):*
> ```Go
> 	req := &monitoringpb.CreateMetricDescriptorRequest{
> 		Name:             projectName,
> 		MetricDescriptor: metric.MetricDescriptor(projectName),
> ```

works for me, seems like a positive change

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L-mojbKhFuAeBqvXsTq:b-m7cp4p)*

knyar:
Sounds good. I've added a new configuration option for this, which can be enabled in the manifest.

---

Review status: 1 of 11 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/stackdriver-nozzle/messages/metric.go, line 70 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mnCUDZ7DwQUzjR--n:-L-q1fw_VpjvcX5AncAG:b-u6dky1) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/messages/metric.go#L70)):*

is this TODO relevant or copy/pasted?
</blockquote></details>

It has been copy/pasted from metric_adapter.go. Should I remove it?

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 79 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kdYDlgkMRrmNNl2AD:-L-q1wwtvo9Mz0ng0esf:b-t3esum) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L79)):*

I think a fixed period would be just fine and more predicable (1? 5? 10 seconds?). Agree with the TTL increase being larger than two metric periods.
</blockquote></details>

As Alex mentioned, there's value in having period decrease for short TTLs, mostly to make tests pass faster. In production it will probably always be capped by `maxExpirePeriod`.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 114 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-he3AaDnOUKH3AZS11:-L-q2QG2jczLakulpzFd:b79jtsc) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L114)):*

0 seems like a potential real value to me. If a counter was 'outOfMemoryErrors' then it very well may emit a cumulative 0 for a long period of time. I think we want to record that value in Stackdriver Monitoring and not hide it.
</blockquote></details>

Very good point, thanks. I've actually changed the condition in the metric sink to check for non-zero time interval, which is what Stackdriver actually expects.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 165 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mqKynAikVOa-wSlEv:-L-q36K3yMn_nIf3b_NK:bn8yz5c) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/nozzle/counter_tracker.go#L165)):*

suggest logging before resetting the value. is it necessary to reset the value? it looks like newCounterData will take care of that if it reappears.
</blockquote></details>

Added logging. Resetting to -1 makes expired counters distinguishable from active ones on the `/debug/vars` HTTP page. I've added a comment clarifying this.

---

*[src/stackdriver-nozzle/nozzle/counter_tracker_test.go, line 69 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kfKTWRHX1XHZFnYcf:-L-q3Y2RqjVuzvf2Xpet:b-ucfknh) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker_test.go#L69)):*

It's totally reasonable and a good clean up. Ginkgo allows you to create custom matchers[1] but it may not be the right fit here.

[1] https://onsi.github.io/gomega/#adding-your-own-matchers
</blockquote></details>

Thanks!

---

*[src/stackdriver-nozzle/stackdriver/metric_adapter.go, line 155 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mrbBA3GzAZbFTt9X7:-L-q3anid2F-LbvVXM5u:b-d9v33m) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/stackdriver/metric_adapter.go#L155)):*

works for me, seems like a positive change
</blockquote></details>

Cool, thanks for the feedback.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L-qIloWg94unnoCc2iG:b-1zbw8w)*

johnsonj:
:lgtm:

---

Reviewed 10 of 10 files at r3.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[jobs/stackdriver-nozzle/spec, line 71 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-rwWKXo8FuLPo0zvtL:-L-rwWKXo8FuLPo0zvtM:bfdhe7k) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/69ecbb10c897850d8d40c7d8e298a79d550c9786/jobs/stackdriver-nozzle/spec#L71)):*
> ```
>     default: cf
> 
>   nozzle.enable_cumulative_counters:
> ```

Just a note- I think it's good to leave this out of the tile.yaml.erb for now. Likely we'll want to just enable this feature once the upstream work lands in PCF.

---

*[src/stackdriver-nozzle/messages/metric.go, line 70 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mnCUDZ7DwQUzjR--n:-L-rvJtUuBywdCyreM7T:b-p674a7) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/messages/metric.go#L70)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

It has been copy/pasted from metric_adapter.go. Should I remove it?
</blockquote></details>

yes please! no clue what it's a TODO - to do :)

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 79 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-kdYDlgkMRrmNNl2AD:-L-rvTKokFkYF4i3cNBx:brlwaxk) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/5ad22e8fa3a7d69d47080b814b16c4039442ad5a/src/stackdriver-nozzle/nozzle/counter_tracker.go#L79)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

As Alex mentioned, there's value in having period decrease for short TTLs, mostly to make tests pass faster. In production it will probably always be capped by `maxExpirePeriod`.
</blockquote></details>

Ah ok thanks! 

---

*[src/stackdriver-nozzle/nozzle/counter_tracker.go, line 165 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mqKynAikVOa-wSlEv:-L-rwOTFUNeg2MJiVHUl:b9nanrb) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/nozzle/counter_tracker.go#L165)):*
<details><summary><i>Previously, knyar (Anton Tolchanov) wrote…</i></summary><blockquote>

Added logging. Resetting to -1 makes expired counters distinguishable from active ones on the `/debug/vars` HTTP page. I've added a comment clarifying this.
</blockquote></details>

thanks! that makes sense

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L-rwkELqlFwsoK6cDfh:bnfp4nl)*

fluffle:
:lgtm: Nice :-)

---

Review status: 10 of 11 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-:-L04Ur3z_SuSLzfHBBPL:b-g6ap26)*

fluffle:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162)*

knyar:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[jobs/stackdriver-nozzle/spec, line 71 at r3](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-rwWKXo8FuLPo0zvtL:-L04XJit3On16I6DJcuA:brxp9sq) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/69ecbb10c897850d8d40c7d8e298a79d550c9786/jobs/stackdriver-nozzle/spec#L71)):*

Just a note- I think it's good to leave this out of the tile.yaml.erb for now. Likely we'll want to just enable this feature once the upstream work lands in PCF.
</blockquote></details>

Sounds good.

---

*[src/stackdriver-nozzle/messages/metric.go, line 70 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162#-L-mnCUDZ7DwQUzjR--n:-L04IazegMQT4UHy5plh:b-896fix) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/09b67bb82bfd4604dfef212a47451891a976c185/src/stackdriver-nozzle/messages/metric.go#L70)):*

yes please! no clue what it's a TODO - to do :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/162)*

knyar:
@johnsonj, I believe this is now ready for you to merge. I don't seem to be able to merge myself because of some pending comments in Reviewable, even though I believe I've addressed all of them.

PR #166, opened 2017-12-12
Add Stackdriver spinner

hustons:
This app can be used to measure the reliability of the stackdriver nozzle

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/166)


src/stackdriver-spinner/main.go
```diff
@@ -0,0 +1,70 @@
+package main
+
+import (
+	"os"
+
```
johnsonj: nit: formatting here. no idea why goland does this

johnsonj: nit: formatting here. no idea why goland does this

src/stackdriver-spinner/README.md
```diff
@@ -0,0 +1,9 @@
+# Log Spinner CF App
+
```
johnsonj: Can you add some documentation about what this app does, the specific metrics it emits, and how it fits into the full pipeline?

It's too early to say "push these buttons and here's your SLO" but providing context will help us stay fresh

src/stackdriver-spinner/README.md
```diff
@@ -0,0 +1,9 @@
...
+## Instructions to run stackdriver-spinner
+1. Add the target GCP project in `stackdriver-tools/src/stackdriver-spinner/manifest.yml`.
```
johnsonj: make these files relative to this directory

src/stackdriver-spinner/README.md
```diff
@@ -0,0 +1,9 @@
+# Log Spinner CF App
```
johnsonj: stackdriver-spinner

src/stackdriver-spinner/main.go
```diff
@@ -0,0 +1,70 @@
...
+	s := session.NewSession(emitter, probe)
+	for {
+		result, err := s.Run(count, waitTime)
+		if err != nil {
+			log.Println(err)
+			continue
+		}
+	logger, err := stackdriver.NewLogger(proj)
+
+	logger.Publish(fmt.Sprintf("GUID: %s - Found: %d - Loss: %.2f \n", result.GUID, result.Found, result.Loss))
```
johnsonj: great place to start! do you plan on following this up with emitting this to stackdriver monitoring or other thoughts around that?

src/stackdriver-spinner/stackdriver/logging_probe_test.go
```diff
@@ -0,0 +1,9 @@
+package stackdriver_test
```
johnsonj: may as well drop this/the suite

src/stackdriver-spinner/main.go
sahilm: The thought here was that we can emit these metrics as regular logs and then use log based metrics for monitoring and alerting. What are your thoughts on this? Would it be better to emit monitoring metrics instead?

- Sahil & Garima

sahilm:
Hey @johnsonj  - 

We've tried to address your feedback and have added a README with a diagram. Let us know what you think and please share any other feedback. Thanks!

Sahil & Garima

src/stackdriver-spinner/main.go
johnsonj: Is it easy to extract the numeric value from log based metrics and graph on that? The final use-case I see here is a graph of '% Logging Loss over time' with alerting thresholds setup by the user (eg, over 10% for 5 minutes).

I think it's good to do both because it's possible to use the Stackdriver Nozzle w/o Stackdriver Monitoring and there is associated cost with using Monitoring/custom metrics.

PR #175, opened 2018-01-04
Per-app HTTP metrics derived from HttpStartStop events.

fluffle:
This will create Stackdriver metrics to track the request rates and response codes of individual PCF apps from the gorouter's HttpStartStop events. It allows the creation of simple SLIs based on e.g. the ratio of 500s.

I intend to use this as a proof-of-concept for per-application SLOs. It's not clear that this functionality *should* be in the nozzle, though if we want proper counters and latency distributions then it kinda has to be. I would appreciate code reviews anyway if you're not too busy :-)

Please don't merge this until we're sure the nozzle is the right place for this code, instead of e.g. having the gorouter track and emit these metrics.

/cc @johnsonj @knyar

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/175)

### DIFF
--- a/jobs/stackdriver-nozzle/spec
+++ b/jobs/stackdriver-nozzle/spec
@@ -73,6 +73,10 @@ properties:
     description: Enable reporting counter events as cumulative Stackdriver metrics. This requires all CounterEvent messages for a given metric to be routed to the same nozzle process (which is the case if you run a single copy of the nozzle).
     default: false
 
+  nozzle.enable_app_http_metrics:
+    description: Enable generation of per-app HTTP metrics from HttpStartStop events.
+    default: false
+
   nozzle.event_filters.blacklist:
     description: |
       Should contain an array of maps with three keys 'sink' (valid values:
--- a/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
+++ b/jobs/stackdriver-nozzle/templates/stackdriver-nozzle-ctl.erb
@@ -39,6 +39,7 @@ case $1 in
     export LOGGING_BATCH_COUNT=<%= p('nozzle.logging_batch_count', '1000') %>
     export LOGGING_BATCH_DURATION=<%= p('nozzle.logging_batch_duration', '30') %>
     export ENABLE_CUMULATIVE_COUNTERS=<%= p('nozzle.enable_cumulative_counters', 'false') %>
+    export ENABLE_APP_HTTP_METRICS=<%= p('nozzle.enable_app_http_metrics', 'false') %>
 
     <% if_p('gcp.project_id') do |prop| %>
     export GCP_PROJECT_ID=<%= prop %>
--- a/src/stackdriver-nozzle/app/builder.go
+++ b/src/stackdriver-nozzle/app/builder.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/telemetry"
 	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/version"
 	"github.com/cloudfoundry/lager"
+	"github.com/cloudfoundry/sonde-go/events"
 )
 
 type App struct {
@@ -80,12 +81,14 @@ func (a *App) newConsumer(ctx context.Context) (nozzle.Nozzle, error) {
 		return nil, err
 	}
 
+	sinks := []nozzle.Sink{}
 	logAdapter := a.newLogAdapter()
 	filteredLogSink, err := nozzle.NewFilterSink(logEvents, lbl, lwl,
 		nozzle.NewLogSink(a.labelMaker, logAdapter, a.c.NewlineToken))
 	if err != nil {
 		return nil, err
 	}
+	sinks = append(sinks, filteredLogSink)
 
 	// Destination for metrics
 	metricAdapter := a.newMetricAdapter()
@@ -101,8 +104,18 @@ func (a *App) newConsumer(ctx context.Context) (nozzle.Nozzle, error) {
 	if err != nil {
 		return nil, err
 	}
+	sinks = append(sinks, filteredMetricSink)
 
-	return nozzle.NewNozzle(a.logger, filteredLogSink, filteredMetricSink), nil
+	if a.c.EnableAppHttpMetrics {
+		httpSink := nozzle.NewHttpSink(a.logger, a.labelMaker)
+		filteredHttpSink, err := nozzle.NewFilterSink([]events.Envelope_EventType{events.Envelope_HttpStartStop}, nil, nil, httpSink)
+		if err != nil {
+			return nil, err
+		}
+		sinks = append(sinks, filteredHttpSink)
+	}
+
+	return nozzle.NewNozzle(a.logger, sinks...), nil
 }
 
 func (a *App) newLogAdapter() stackdriver.LogAdapter {
--- a/src/stackdriver-nozzle/config/config.go
+++ b/src/stackdriver-nozzle/config/config.go
@@ -90,6 +90,9 @@ type Config struct {
 	// requires deterministic routing of CounterEvents to nozzles (i.e. CounterEvent messages for a particular metric MUST
 	// always be routed to the same nozzle process); the easiest way to achieve that is to run a single copy of the nozzle.
 	EnableCumulativeCounters bool `envconfig:"enable_cumulative_counters"`
+	// If enabled, the Nozzle will derive per-application HTTP metrics from
+	// HttpStartStop events and export them as counters to Stackdriver.
+	EnableAppHttpMetrics bool `envconfig:"enable_app_http_metrics"`
 	// Expire internal counter state if a given counter has not been seen for this many seconds.
 	CounterTrackerTTL int `envconfig:"counter_tracker_ttl" default:"130"`
 
--- a/src/stackdriver-nozzle/messages/metric.go
+++ b/src/stackdriver-nozzle/messages/metric.go
@@ -109,15 +109,26 @@ func (m *Metric) Hash() string {
 	var b bytes.Buffer
 
 	b.Write([]byte(m.Name))
+	if len(m.Labels) > 0 {
+		b.WriteByte(',')
+		b.WriteString(Flatten(m.Labels))
+	}
+	return b.String()
+}
 
-	// Extract label keys to a slice and sort it
-	keys := make([]string, 0, len(m.Labels))
-	for k := range m.Labels {
+// Flatten serializes a set of label keys and values.
+func Flatten(labels map[string]string) string {
+	keys := make([]string, 0, len(labels))
+	for k := range labels {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
-	for _, k := range keys {
-		b.Write([]byte(fmt.Sprintf(",%s='%s'", k, m.Labels[k])))
+	buf := &bytes.Buffer{}
+	for i, k := range keys {
+		buf.WriteString(fmt.Sprintf("%q=%q", k, labels[k]))
+		if i+1 < len(keys) {
+			buf.WriteByte(',')
+		}
 	}
-	return b.String()
+	return buf.String()
 }
--- a/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go
+++ b/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package nozzle
+
+import (
+	"fmt"
+
+	"github.com/cloudfoundry-community/stackdriver-tools/src/stackdriver-nozzle/telemetry"
+	"github.com/cloudfoundry/lager"
+	"github.com/cloudfoundry/sonde-go/events"
+)
+
+const httpPrefix telemetry.MetricPrefix = "app-http"
+
+var (
+	requestCount *telemetry.CounterMap
+	responseCode *telemetry.CounterMap
+
+	defaultLabels = []string{"job", "index", "applicationPath", "instanceIndex"}
+)
+
+func init() {
+	requestCount = telemetry.NewCounterMap(httpPrefix, "request_count",
+		defaultLabels...)
+	responseCode = telemetry.NewCounterMap(httpPrefix, "response_code",
+		append(defaultLabels, "code")...)
+}
+
+type httpSink struct {
+	logger     lager.Logger
+	labelMaker LabelMaker
+}
+
+// NewHttpSink returns a Sink that can receive sonde HttpStartStop events
+// and generate per-application HTTP metrics from them.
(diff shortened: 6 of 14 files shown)


johnsonj:
I took a high level view of this change and it looks good to me. I don't know where else this sort of code would live. The alternative to this is some sort of 'CalcualtedMetricsSink' that has handlers for various event types. That's probably overkill for the scope of change but worth considering if you think these calculated metrics are going to grow.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/stackdriver/telemetry_sink.go, line 80 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L21TRwt5CJQ4z1Fvhh3:-L21TRwt5CJQ4z1Fvhh4:b-i0en6m) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/60897e829174561baac32eef84538d597f407ee6/src/stackdriver-nozzle/stackdriver/telemetry_sink.go#L80)):*
> ```Go
> 		client:      client,
> 		projectPath: fmt.Sprintf("projects/%s", projectID),
> 		labels:      map[string]string{"subscription_id": subscriptionId, "foundation": foundation},
> ```

Can you extract this commit to its own PR?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L22w7_1xjmJDb99II-g:b6xtbjz)*

fluffle:
Thanks for taking a look!

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/stackdriver-nozzle/stackdriver/telemetry_sink.go, line 80 at r1](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L21TRwt5CJQ4z1Fvhh3:-L25Vyw_UXs2gPMipNuj:b-8n3lc8) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/60897e829174561baac32eef84538d597f407ee6/src/stackdriver-nozzle/stackdriver/telemetry_sink.go#L80)):*

Can you extract this commit to its own PR?
</blockquote></details>

Sure! I totally forgot about that commit, pretty sure I intended to do that anyway :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L25WL6YzeYitoybglYc:bubxju8)*

johnsonj:
Let me know when you want me to take a full look at this one

fluffle:
Thanks! It'll be some time next week, I'm trying to get the automon-side graphs in place too so I can demonstrate utility (and verify this is doing the right thing).

fluffle:
I think this is pretty much good to go, and I think we'd like to get this code into the nozzle at least for the time being because per-App HTTP stats and SLOs have some value for us.

johnsonj:
:lgtm: Please address comments. Feel free to merge after doing so. Let me know if you want me to take a look at a revision. 

---

Reviewed 3 of 9 files at r1, 11 of 11 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/stackdriver-nozzle/app/builder.go, line 112 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Kirx8KEuTBVS1Umt:-L3-Kirx8KEuTBVS1Umu:bgfw6ya) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/app/builder.go#L112)):*
> ```Go
> 			return nil, err
> 		}
> 		return nozzle.NewNozzle(a.logger, filteredLogSink, filteredMetricSink, filteredHttpSink), nil
> ```

instead of introducing multiple NewNozzle calls, can you create a []Sink, populate it throughout this method, and return nozzle.NewNozzle(a.logger, sinks...)

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink.go, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-L-sqd2ZREhvkBhgs:-L3-L-sqd2ZREhvkBhgt:b-1d0p0) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go#L38)):*
> ```Go
> 
> func init() {
> 	// The telemetry sink handles the "foundation" label for us \o/
> ```

I don't think this is really needed

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink.go, line 67 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-LB4as-3NK3-aIsx0:-L3-LB4as-3NK3-aIsx1:boicxyw) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go#L67)):*
> ```Go
> 
> 	labels := sink.labelMaker.MetricLabels(envelope, false)
> 	if envelope.GetJob() != "router" || labels["applicationPath"] == "" {
> ```

check for `envelope.GetJob() != "router"` above, before generating labels 

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink_test.go, line 149 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-LmsFTKw7GpqwAC-c:-L3-LmsFTKw7GpqwAC-d:b-pn36qg) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink_test.go#L149)):*
> ```Go
> 		// AppOne has 2 instances.
> 		// The first serves 20 "200 OK" responses.
> 		receive(testApps[0].Events(20, 200, 0))
> ```

could this be better represented as a table test? 

---

*[src/stackdriver-nozzle/nozzle/label_maker.go, line 127 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Lygh43bVSHIG3eTl:-L3-Lygh43bVSHIG3eTm:buzlm9g) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/label_maker.go#L127)):*
> ```Go
> }
> 
> // Separated out for use in tests.
> ```

I'd drop this comment. There isn't a risk of someone folding the code back in since it would break the build.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 69 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Mb2qAqzUQbAOH_9q:-L3-Mb2qAqzUQbAOH_9r:bht2kw9) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/telemetry/counters.go#L69)):*
> ```Go
> 	existing := cm.Get(key)
> 	if existing == nil {
> 		// TODO(fluffle): this might race?
> ```

can you resolve this? 

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 92 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-MhKnDh6Xd_QC73sd:-L3-MhKnDh6Xd_QC73se:bht2kw9) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/telemetry/counters.go#L92)):*
> ```Go
> // MapKey serializes the set of label keys and values for use as
> // the key in an expvar.Map.
> // TODO(fluffle): Dedupe this with messages.Metric.Hash()?
> ```

can you resolve this? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L3-Ms2gG0w3KEtoLB1v:b-iqxwyo)*

fluffle:
Thanks for the review! I should probably build a new version and do some manual integration testing of the varz output of counters etc. before we merge this -- I've only run the unit tests today.

---

Review status: 7 of 15 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/stackdriver-nozzle/app/builder.go, line 112 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Kirx8KEuTBVS1Umt:-L3E7-nplTGZTOPEF153:b-dtkds3) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/app/builder.go#L112)):*

instead of introducing multiple NewNozzle calls, can you create a []Sink, populate it throughout this method, and return nozzle.NewNozzle(a.logger, sinks...)
</blockquote></details>

Done, thanks!

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink.go, line 38 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-L-sqd2ZREhvkBhgs:-L3E7yeVUlM1i3zGUnUW:b-g2z0ca) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go#L38)):*

I don't think this is really needed
</blockquote></details>

Ack, removed.

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink.go, line 67 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-LB4as-3NK3-aIsx0:-L3E7HSGaIPNKvmf92J0:bdva0w2) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink.go#L67)):*

check for `envelope.GetJob() != "router"` above, before generating labels 
</blockquote></details>

It turns out that this plays badly with isolation segments, because the "job" label is used to distinguish between routers in different segments. 

https://docs.cloudfoundry.org/adminguide/routing-is.html 

So I nuked the test and added "job" to the list of default labels -- which I had originally but changed because it looked like "job" was a useless label :-(

---

*[src/stackdriver-nozzle/nozzle/httpstartstop_sink_test.go, line 149 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-LmsFTKw7GpqwAC-c:-L3EA6wDMmCVShW_WktL:b1k87lr) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/httpstartstop_sink_test.go#L149)):*

could this be better represented as a table test? 
</blockquote></details>

I tried!

Because there's a number of dimensions and a 1:many mapping between the input data and the outputs to test, I couldn't come up with a neat way of translating it into a table of inputs and outputs with generic testing code. If you have any smart ideas I'd love to hear them :-(

---

*[src/stackdriver-nozzle/nozzle/label_maker.go, line 127 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Lygh43bVSHIG3eTl:-L3EB7X71XFvEB4Uujia:b-g2z0ca) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/nozzle/label_maker.go#L127)):*

I'd drop this comment. There isn't a risk of someone folding the code back in since it would break the build.
</blockquote></details>

Ack, removed.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 69 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Mb2qAqzUQbAOH_9q:-L3EGAFF3VlyRNDLbXZy:bcss6cs) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/telemetry/counters.go#L69)):*

can you resolve this? 
</blockquote></details>

I'm categorically awful at figuring out races. In this case I think it can, because while the Get and Set operations are atomic, the Set is not a CompareAndSwap, so it is perfectly possible for two goroutines to both Get a nil value and then have the second Set overwrite the first.

However I don't think the consequences of this are particularly bad unless the first Counter returned is stored for a long time elsewhere and incremented repeatedly, while the second is not. The "easy" fix is to just lock the map while getting/setting. I should just do this, I guess.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 92 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-MhKnDh6Xd_QC73sd:-L3EBIT6rzCETr0fUmbg:b-w46yll) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/telemetry/counters.go#L92)):*

can you resolve this? 
</blockquote></details>

Arghg, caught out by the idle thoughts of past me. Past me is a terrible person.

MapKey and Hash are behaviourally quite similar. MapKey needs to be safe for use in JSON hence the %q formatting. (I think this is true of Hash now, too, actually, so that's a good reason to merge them somehow.)  But I stumbled over the question of where to put the shared function. Package messages is probably the right place, since it's trivial and at the bottom of the dep tree. This will also change varz formatting for counters somewhat, hopefully that will be ok, I should possibly do more due diligence here next week before merging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L3EKz-JYC0g2igGUSsQ:b-hizign)*

fluffle:
Everything looks good. I nuked the response class metric again, because I think I can get Monarch / Stackdriver to do the class aggregation for me. I don't like the multiple layers of escaping in varz output and It's probably unneccessary in 99% of cases, but I think it's probably better to be safe, given that the chances of anyone actually consuming those varz are also tiny.

I can't merge until you click some buttons in reviewable to resolve things. Yay :-)

---

Review status: 6 of 14 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L3YRoHQMgUNGh0986o5:b-xpj43w)*

johnsonj:
Ohh process.. I'll try to disable that requirement

---

Reviewed 8 of 9 files at r3.

---

*[src/stackdriver-nozzle/telemetry/counters.go, line 69 at r2](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-L3-Mb2qAqzUQbAOH_9q:-L3ZHMHyW1Av_VAwNqkz:bb3zgha) ([raw file](https://github.com/cloudfoundry-community/stackdriver-tools/blob/3424d80d776328d37ae8de9ca0d0fe689b775fd7/src/stackdriver-nozzle/telemetry/counters.go#L69)):*

I'm categorically awful at figuring out races. In this case I think it can, because while the Get and Set operations are atomic, the Set is not a CompareAndSwap, so it is perfectly possible for two goroutines to both Get a nil value and then have the second Set overwrite the first.

However I don't think the consequences of this are particularly bad unless the first Counter returned is stored for a long time elsewhere and incremented repeatedly, while the second is not. The "easy" fix is to just lock the map while getting/setting. I should just do this, I guess.
</blockquote></details>

Looks goo to me!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cloudfoundry-community/stackdriver-tools/175#-:-L3ZHZ0JOCbUM7ByVTpt:b-jg8xrt)*

Issue #209, opened 2018-09-18
Log Message is Duplicated


pivotalsquid:
We're using the Stackdriver Nozzle to get logs from our application running on CloudFoundry into Stackdriver.

Viewing the logs coming straight out of PCF, I get a structure like the following:

```
{
  "timestamp": "…",
  "source_id": "…",
  "instance_id": "…",
  "deprecated_tags": {…},
  "tags": {
    "deployment": "…",
    "index": "…",
    "ip": "…",
    "job": "…",
    "origin": "…",
    "source_type": "…"
  },
  "log": {
    "payload": "Base64encoded..."
  }
}
```

Viewing this same log in Stackdriver, I see this structure:
```
{
 insertId:  "…"  
 jsonPayload: {
  deployment:  "…"   
  eventType:  "…"   
  index:  "…"   
  ip:  "…"   
  job:  "…"   
  logMessage: {
   app_id:  "…"    
   message:  "Base64encoded..."    
   message_type:  "…"    
   source_instance:  "…"    
   source_type:  "…"    
   timestamp:  …    
  }
  message:  "Base64encoded..."   
  origin:  "…"   
  serviceContext: {…}   
  tags: {
   source_id:  "…"    
  }
  timestamp:  …   
 }
 labels: {…}  
 logName:  "…"  
 receiveTimestamp:  "…"  
 resource: {…}  
 timestamp:  "…"  
}
```

Notice how the `log.payload` from PCF is duplicated in 2 places by the time it gets into Stackdriver, In the `jsonPayload.logMessage.message` and `jsonPayload.message` fields.

Is this intentional? If so, why? It has caused a bit of confusion - which `message` field should we be using? We are also wondering that this duplication may be increasing our Stackdriver and BigQuery storage costs.

mattysweeps:
This was merged to master via #211

PR #211, opened 2018-10-15
Remove the duplication of the log message

pivotalsquid:
The message field in the logMessage payload was being duplicated into the payload.message field where Stackdriver is expecting to find it. There is no reason to have this information duplicated. It increases storage costs and is confusing for people to see the same information twice in search results.

I wrote this pull request to satisfy issue [#209](https://github.com/cloudfoundry-community/stackdriver-tools/issues/209)

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/211)

### DIFF
--- a/src/stackdriver-nozzle/nozzle/log_sink.go
+++ b/src/stackdriver-nozzle/nozzle/log_sink.go
@@ -84,11 +84,14 @@ func (ls *logSink) parseEnvelope(envelope *events.Envelope) messages.Log {
 			// to remain consistent with the protobuf.
 			logMessageMap["message_type"] = logMessage.GetMessageType().String()
 			severity = parseSeverity(logMessage.GetMessageType())
-			logMessageMap["message"] = message
-			payload["logMessage"] = logMessageMap
 
-			// Duplicate the message payload where stackdriver expects it
+			// Put the message payload where stackdriver expects it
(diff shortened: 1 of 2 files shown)


src/stackdriver-nozzle/nozzle/log_sink.go
```diff
@@ -84,10 +84,10 @@ func (ls *logSink) parseEnvelope(envelope *events.Envelope) messages.Log {
 			// to remain consistent with the protobuf.
 			logMessageMap["message_type"] = logMessage.GetMessageType().String()
 			severity = parseSeverity(logMessage.GetMessageType())
-			logMessageMap["message"] = message
+			delete(logMessageMap, "message")
```
mattysweeps: @pivotalsquid Can you do some rearranging? Can you move the `payload["message"] = message` line before the delete?

Then, add a comment to the delete line which states the delete is removing the duplicate message 

This would make the reasoning for the delete more obvious.

pivotalsquid: Sure, I can make that change.

pivotalsquid: Change complete.

PR #219, opened 2018-12-07
Simplify tile configuration

mattysweeps:
* Generate an x509 certificate for RLP's mutual TLS
* Derive the root Certificate Authority
* Add the BOSH DNS hostname for the Reverse Log Proxy as a default value
* Derive the API URL

I've tested this works with the large footprint. I'm currently working on testing this with the small footprint. However, I ran into issues with the small footprint before these changes, so I'm skeptical. We'll see.

I also updated some of the tile building scripts to make it easier for me to develop.

---
This change is [Reviewable](https://reviewable.io/reviews/cloudfoundry-community/stackdriver-tools/219)

### DIFF
--- a/tile.yml.erb
+++ b/tile.yml.erb
@@ -20,6 +20,8 @@ packages:
     templates:
     - name: stackdriver-nozzle
       release: stackdriver-tools
+      consumes:
+        reverse_log_proxy: {from: reverse_log_proxy, deployment: "(( ..cf.deployment_name ))"}
     memory: 512
     ephemeral_disk: 4096
     cpu: 2
@@ -29,7 +31,6 @@ packages:
     instances: 1
     properties:
       rlp:
-        address: (( .properties.rlp_address.value ))
         ca_cert: (( $ops_manager.ca_certificate ))
         cert: (( .properties.rlp_creds.cert_pem ))
         key: (( .properties.rlp_creds.private_key_pem ))
@@ -64,11 +65,6 @@ forms:
   label: Nozzle Configuration
   description: Configure access properties for the Stackdriver Nozzle here
   properties:
-  - name: rlp_address
-    type: string
-    label: Reverse Log Proxy host and port
-    description: Address and port of the reverse log proxy. The default value works if BOSH DNS is enabled. Otherwise, set this to the IP address of a reverse log proxy VM.
-    default: loggregator-trafficcontroller.service.cf.internal:8082
   - name: rlp_creds
     type: rsa_cert_credentials
     label: rlp_creds
--- a/tile.yml.erb
+++ b/tile.yml.erb
@@ -30,11 +30,11 @@ packages:
     properties:
       rlp:
         address: (( .properties.rlp_address.value ))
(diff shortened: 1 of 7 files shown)


tile.yml.erb
```diff
@@ -66,20 +66,16 @@ forms:
...
-    default: (( $runtime.system_api_url ))
+    label: Reverse Log Proxy host and port
+    description: Address and port of the reverse log proxy. The default value works if BOSH DNS is enabled. Otherwise, set this to the IP address of a reverse log proxy VM.
+    default: loggregator-trafficcontroller.service.cf.internal:8082
+  - name: rlp_creds
+    type: rsa_cert_credentials
+    label: rlp_creds
+    configurable: false
+    default:
+      domains:
```
evandbrown: What does this config key do?

tile.yml.erb
```diff
@@ -66,20 +66,16 @@ forms:
...
-  - name: rlp_key
-    type:  text
-    label: The TLS key for the Reverse Log Proxy, found in the same place as the cert
-  - name: firehose_endpoint
-    type: string
-    label: Cloud Foundry API Endpoint
-    default: (( $runtime.system_api_url ))
+    label: Reverse Log Proxy host and port
+    description: Address and port of the reverse log proxy. The default value works if BOSH DNS is enabled. Otherwise, set this to the IP address of a reverse log proxy VM.
+    default: loggregator-trafficcontroller.service.cf.internal:8082
```
evandbrown: I wonder if it's possible to discover this via a [BOSH link](https://bosh.io/docs/links/)?

tile.yml.erb
mattysweeps: Sets the CN and the X509v3 Subject Alternative Name of the certificate.

This key isn't necessary for the nozzle to work. However, without it the certificate defaults its CN and X509v3 Subject Alternative Name to the name of the property key: `rlp_creds`. I thought that name would be confusing if someone inspected the certificate, so I made it `stackdriver-nozzle` instead.

tile.yml.erb
mattysweeps: Should be, I'll investigate tomorrow and update this PR

mattysweeps:
@evandbrown updated

tile.yml.erb
```diff
@@ -29,12 +31,11 @@ packages:
...
-        address: (( .properties.rlp_address.value ))
-        ca_cert: (( .properties.rlp_cacert.value ))
-        cert: (( .properties.rlp_cert.value ))
-        key: (( .properties.rlp_key.value ))
+        ca_cert: (( $ops_manager.ca_certificate ))
+        cert: (( .properties.rlp_creds.cert_pem ))
+        key: (( .properties.rlp_creds.private_key_pem ))
       firehose:
-        endpoint: (( .properties.firehose_endpoint.value ))
+        endpoint: (( $runtime.system_api_url ))
```
mattysweeps: I removed the firehose_endpoint property from the tile. I assumed that the API targets will be the same as the runtime deploying the tile. 

If this is something we should keep open for configuration, let me know.

PR #232, opened 2019-01-17
Update tile form description for service account

mattysweeps:


tile.yml.erb
```diff
@@ -98,7 +98,7 @@ forms:
   - name: service_account
     type: text
     label: Service Account Credentials
-    description: JSON string credentials for a Service Account with logging.logWriter and logging.configWriter permissions
+    description: "JSON string credentials for the Service Account used by the nozzle. Adding the 'Logging Admin' and 'Monitoring Admin' is a superset of the permissions needed."
```
josephlewis42: JSON string credentials for the service account used by the nozzle. The service account should have both the `logging.admin` and `monitoring.admin` roles.

mattysweeps: @josephlewis42 logging.logWriter and logging.configWriter are rules not permissions. Also they together do not cover all the required permissions. So they initial comment was wrong on many counts.

I can apply your first suggestion

